PR #100, opened 2014-03-10
Possible keybinding to switch to previous workspace.

robertek:
This is implementation of new keybinding handler. It allows to switch to previously selected workspece. It is very useful for example for fast switching between two workspaces.

### DIFF
--- a/src/50-marco-global-key.xml.in
+++ b/src/50-marco-global-key.xml.in
@@ -127,5 +127,12 @@
 	schema="org.mate.Marco.general"
 	comparison="gt" />
 
+	<KeyListEntry name="switch-to-workspace-prev"
+	_description="Switch to previously selected workspace"
+	value="1"
+	key="num-workspaces"
+	schema="org.mate.Marco.general"
+	comparison="gt" />
+
 </KeyListEntries>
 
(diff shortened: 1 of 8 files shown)


robertek:
I have found that. It is caused somehow by the order of keybind in all-keybindings.h
This is wired and still did not found a root cause, because this is then parsed in few files as xml or data structure. The following patch fix this but Im trying to find out better way. I will create new pull request whit the whole patch then.

diff --git a/src/include/all-keybindings.h b/src/include/all-keybindings.h
index 6707a21..50532c4 100644
--- a/src/include/all-keybindings.h
+++ b/src/include/all-keybindings.h
@@ -111,9 +111,6 @@ keybind (switch-to-workspace-up, handle_switch_to_workspace,
 keybind (switch-to-workspace-down, handle_switch_to_workspace,
          META_MOTION_DOWN, 0)
 
-keybind (switch-to-workspace-prev, handle_switch_to_workspace,
-         META_MOTION_PREV, 0)
-
 /***********************************/
 
 /* The ones which have inverses.  These can't be bound to any keystroke
@@ -275,5 +272,8 @@ keybind (move-to-side-w, handle_move_to_side_w, 0,
 keybind (move-to-center, handle_move_to_center, 0,
         BINDING_PER_WINDOW)
 
+keybind (switch-to-workspace-prev, handle_switch_to_workspace,
+         META_MOTION_PREV, 0)
+
 /* eof all-keybindings.h */

PR #113, opened 2014-06-22
Repaint area of deleted shadowed windows (fixes #27)

### LINKED ISSUE #27: Black shadow outline remains after network connection tooltip
I've experienced this issue in all of the MATE installations I've done so far (all of them with Linux Mint), on different architectures, hardware  profiles, and MATE versions (from 1.2 onwards).

Simply put, when I hover the pointer over the network connection icon at the bottom right of the screen, a small pop-up appears with some details about the current active connection. This pop-up is standard behaviour on all of the plugins in the desktop bar, as far as I can tell. But the network connection plugin has an unexpected behaviour, in that the shadow that accompanies the little pop-up remains in the screen after the pop-up is gone. See the following screencaps for an example:

pop-up message with outline present: http://s10.postimage.org/9c8z6f8uh/Screenshot4.png
pop-up message gone, the outline remains: http://postimage.org/image/aryzv3gff/

I'm pretty ure it's a Marco thing, because I briefly tried Compiz and this didn't happen.

bl0ckeduser:
Hi, this is a fix for the bug where the network manager applet tooltip's shadow didn't disappear after the tooltip disappear.

The bug also happened for pretty much any tooltip from any GTK3 application, e.g. brasero

The fix code is checking if a window with a shadow has just been deleted, and if so, it is repainting the area of the screen that was occupied by the window.

Best regards :)

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -1544,6 +1544,25 @@ free_win (MetaCompWindow *cw,
   MetaDisplay *display = meta_screen_get_display (cw->screen);
   Display *xdisplay = meta_display_get_xdisplay (display);
   MetaCompScreen *info = meta_screen_get_compositor_data (cw->screen);
+  XRectangle r;
+  XserverRegion region;
+
+  /*
+   * If we are deleting a window with a compositor shadow,
+   * trigger a repaint of the area it was covering.
+   * This fixes an issue where the shadow beneath a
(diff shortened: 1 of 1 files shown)


flexiondotorg:
@stefano-k Please review this for inclusion in MATE 1.10

infirit:
I just merged this as I got this again recently. It only happens very sporadically though..

Issue #97, opened 2014-03-08
Make the reduced resources mode more pleasurable for the eyes


ikem-krueger:
Enable "reduced resources" and you see a wireframe when you move a window.

What I like to see is only the outer frame of the window, not the inner "wires".

monsta:
Wow, the mentioned PR (https://github.com/mate-desktop/marco/issues/122) removes a lot of code. So do we need it?

flexiondotorg:
The MATE team have decided that we don't wish to make this change.

PR #122, opened 2014-06-29
draw only outer frames in "reduced resource" mode

dnk:
fix for https://github.com/mate-desktop/marco/issues/97


flexiondotorg:
@clefebvre @monsta @raveit65 Thoughts on this please?

To me this looks neater, removes code which is a good thing IMO. But will it make merging from Metacity in the future more complicated?

ikem-krueger:
Thank you dnk for making that commit. I safe them for later reference.

Issue #45, opened 2013-04-18
Panels go completely transparent with compositing


infirit:
For some reason when I start calibre (ebook manager) sometimes on startup it makes all panels transparent. The panels are still there and can be interacted.

When I turn compositing off and then on it is back to normal. Or restarting marco has the same affect.

The programs I had running at the time are:
virtualbox running a vm
firefox
hexchat
thunderbird
a couple of mate-terminals

The only thing I see in .xsession-error is the following:
Windowmanager waarschuwing:Ongeldig WM_TRANSIENT_FOR venster 0x4200004 opgegeven voor 0x4200002 (calibre-gu).

The error happens always but it does not make the panels transparent every time.

paulguy:
This needs to be reopened because it's still broken.  I don't know what triggers it but it happened to me and now it won't come back no matter what I do.  I've tried mate-panel --replace, mate-panel --reset, then --replace.  I've tried killing it then reopening it and it just won't come back.  It takes up the space and I can right click on it to bring up menus.  I get this line on the terminal repeatedly:
`(mate-panel:15694): GLib-CRITICAL **: g_hash_table_remove_internal: assertion 'hash_table != NULL' failed`

I am not using the compositor.

4nanook:
I'm experiencing this with mxlinux.

lukefromdc:
I have not seen this sort of thing on Debian Unstable at all, even when I had a temporary test install of Calibre maybe a month or so ago for testing a tray issue.  Anyway, try these tests:

1:
Are you using the system background, a user set color background, or a user set image background? These each use different code. If one is giving you trouble, right-click on any empty space on the panel, click "properties" and in the dialog that comes up open the "background" tab and try a different background. Then let us know what happened.

2: Try setting Calbre not to set a tray icon, and see what happens If this makes a difference then try the next test:

3: Try opening dconf-editor, going to org>mate>panel>general and try both TRUE and FALSE for "enable sni support" and see what happens.

PR #161, opened 2015-01-03
Update compositor-xrender.c

darkshram:
Fixes panels go completely transparent with compositing (issue #45). Backport from https://github.com/mate-desktop/marco/pull/107 to fix https://github.com/mate-desktop/marco/issues/45


infirit:
We do not maintain 1.6 any-more as it is obsolete. The current supported version is 1.8 which has this fix already.

PR #162, opened 2015-01-07
show tile preview with external compositor

dnk:
current implementation of tile preview depends on buildin compositor. 
ie when builtin compositor is disabled, tile preview always display only borders.

with this patch tile preview works well with external compositor( for example with "compton")

### DIFF
--- a/src/core/screen-private.h
+++ b/src/core/screen-private.h
@@ -115,6 +115,7 @@ struct _MetaScreen
 #endif
 
 #ifdef HAVE_COMPOSITE_EXTENSIONS
+  Atom wm_cm_atom;
   Window wm_cm_selection_window;
   guint32 wm_cm_timestamp;
 #endif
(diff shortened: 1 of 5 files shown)


infirit:
This works but it does not handle starting and stopping of the compositor. If the compositor is not running when marco starts then we do not have the composited preview. The reverse also applies that when the compositor is stopped we still get the composited preview.

infirit:
I think it will be better to change the code to listen to the ``compositor-changed`` signal and then update the ``compositor`` variable. This has the added benefit that the other compositing enhancements like alt-tab preview will also "just work".

infirit:
I just noticed the tiling code does not notice the compositor switching on<->off at all for me..

infirit:
testing this with compton and works nicely :smile:

infirit:
Merged, thanks.

monsta:
@flexiondotorg: I think this can be used as a patch for 1.8.x to fix https://bugs.launchpad.net/bugs/1406645.

Issue #158, opened 2014-12-30
Tile preview graphics glitch


adamjsmith:
Steps to reproduce bug:

1. Snap a window to the left of the screen
2. Snap a window to the right of the screen
3. Before the tile preview appears the background of the left of the screen appears very briefly (a fraction of a second) at the right of the screen
4. Snap a window to the left of the screen
5. A similar graphics glitch is observed on the left of the screen (i.e. the right hand side background is flashed on the left of the screen).

This is reproducable under ubuntu-mate 12.04 (precise) and 14.04 (trusty).

Marco version: 1.8.2+dfsg1-4~trusty1

This is on a big-endian machine (PowerPC), but I think the slowness of the machine is more significant.

Launchpad bug: https://bugs.launchpad.net/ubuntu/+source/marco/+bug/1406344

NiceandGently:
@stefano-k 
@flexiondotorg 
commit is in marco-1.10.1, can be closed.

PR #164, opened 2015-01-14
tile preview: invalidate window before showing

dnk:
fix for https://github.com/mate-desktop/marco/issues/158

### DIFF
--- a/src/ui/tile-preview.c
+++ b/src/ui/tile-preview.c
@@ -271,7 +271,6 @@ meta_tile_preview_show (MetaTilePreview *preview,
       && preview->tile_rect.height == tile_rect->height)
     return; /* nothing to do */
 
-  gtk_widget_show (preview->preview_window);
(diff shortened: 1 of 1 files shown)


infirit:
I can still reproduce this after applying this patch. What I do is move it from side to side multiple times, it sometimes takes a few tries but eventually I get a flicker.

NiceandGently:
I can't reproduce this glitches with mate gtk2 and gtk3 builds in fedora 22 (mate-1.10).
Note, i use this PR and the ' adding keybindings for tiling' PR, since several month for mate-1.8/10 builds for official fedora, and have no reports about it.

flexiondotorg:
Thanks for your contribution.

monsta:
@flexiondotorg: I guess https://github.com/mate-desktop/marco/issues/158 (and LP report that's mentioned there) can be closed now?

PR #188, opened 2015-04-08
[1.8] prefs: make workspace name actually change

monsta:
now workspace names actually change on-the-fly when you change them in dconf-editor.
and this works with GLib 2.43 as well.


flexiondotorg:
Thanks for your contribution.

PR #190, opened 2015-04-09
[1.8] prefs: make keybindings change notifications work with GLib >= 2.43

monsta:
this is achieved by using the same GSettings instance for listening
and reading data. it's an additional fix for the issue with GLib >= 2.43,
https://git.gnome.org/browse/glib/commit/?id=8ff5668a458344da22d30491e3ce726d861b3619


flexiondotorg:
Thanks for your contribution.

PR #191, opened 2015-04-17
Remove information about non-existent tool run-marco.sh in HACKING

GiedriusS:
run-marco.sh doesn't exist so remove information about it from the
HACKING file


flexiondotorg:
Thanks for your contribution.

PR #198, opened 2015-06-09
fix uninitialized data field for sending _NET_WM_SYNC_REQUEST event

sgerd:
If this field is left uninitialized, a client may mistake Marco to support more than the single update request counter described in EWMH 1.5.


flexiondotorg:
Thanks for your contribution.

PR #203, opened 2015-07-11
Reverting center-new-windows default to false. 

bjlaur:
This is changed to match the description of the setting. In addition, the behavior to open new windows on the top left of the screen has been consistent for over a decade. I believe it's counter-productive to change this now.

### DIFF
--- a/src/org.mate.marco.gschema.xml
+++ b/src/org.mate.marco.gschema.xml
@@ -162,7 +162,7 @@
       <description>Some applications disregard specifications in ways that result in window manager misfeatures. This option puts Marco in a rigorously correct mode, which gives a more consistent user interface, provided one does not need to run any misbehaving applications.</description>
     </key>
     <key name="center-new-windows" type="b">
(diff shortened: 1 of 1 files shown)


clefebvre:
Ok, we need opinions on this please.

-1 for me (I find it much neater. Monitors are much wider than before too and I rarely map multiple windows at the same time... so I'd rather look in the center than on the side, and I've no need for cascading).

flexiondotorg:
I have a gschema override in Ubuntu MATE to use smart positioning, so don't really mind what the default is.

flexiondotorg:
@raveit65 Please let us know your thoughts on this.

raveit65:
@flexiondotorg 
For fedora i use use a override too which set it to false :)

clefebvre:
Was merged by mistake.
Reverted by http://git.io/vnFTR after reviewing votes (monsta and I against, raveit in favour, flexion absention).

Issue #205, opened 2015-07-17
marco (gtk3) crashes with some metacity themes


NiceandGently:
opening mate-appearance-properties causes a abrt alarm which pointed to marco.
stacktrace:
```
Thread 1 (Thread 0x7f46d0dd8a00 (LWP 2339)):
#0  0x00007f46cb593a98 in __GI___libc_sigaction (sig=2339, act=0x923, oact=0x6) at ../sysdeps/unix/sysv/linux/x86_64/sigaction.c:54
        result = <optimized out>
        kact = {k_sa_handler = 0x7f46d0df4840, sa_flags = 140725531631552, sa_restorer = 0x100000000, sa_mask = {__val = {140725531631536, 252833149, 139942045244075, 4294967295, 139942036386552, 139942035873096, 21543808, 117, 139942048547680, 140725531631920, 139942081571520, 18540032, 139942036068138, 32, 0, 0}}}
        koact = {k_sa_handler = 0x0, sa_flags = 0, sa_restorer = 0x0, sa_mask = {__val = {0, 0, 0, 0, 0, 0, 0, 0, 0, 21543808, 117, 139942048547680, 139942126636496, 117, 139942048547680, 140725531631920}}}
#1  0x00007f46cb5e2e9c in __GI___libc_free (ptr=0x85, str=0x7f46cb6e9808 "munmap_chunk(): invalid pointer", action=21543808) at malloc.c:4973
        buf = "\200\273H\001\000\000\000\000\350D\337\320F\177\000\000\n"
        cp = <optimized out>
        p = 0x75
        size = <optimized out>
        block = <optimized out>
        total_size = <optimized out>
        ar_ptr = <optimized out>
        p = 0x75
        hook = <optimized out>
#2  0x00007f46cb5e2e9c in __GI___libc_free (p=0x75) at malloc.c:2823
        p = 0x75
        size = <optimized out>
        block = <optimized out>
        total_size = <optimized out>
        ar_ptr = <optimized out>
        p = 0x75
        hook = <optimized out>
#3  0x00007f46cb5e2e9c in __GI___libc_free (mem=0x85) at malloc.c:2948
        ar_ptr = <optimized out>
        p = 0x75
        hook = <optimized out>
#4  0x00007f46ce0f698d in  () at /usr/lib64/libmarco-private.so.0
#5  0x00007f46ce0f90a8 in  () at /usr/lib64/libmarco-private.so.0
#6  0x0000000000000523 in  ()
[...]

PR #206, opened 2015-07-17
theme: Remove outdated assertion

cygwinports:
The theme state used to use GtkStateType, but was ported over to GtkStateFlags,
leaving behind a broken assertion that fails when using certain Metacity
themes, for example Nodoka.

https://bugzilla.gnome.org/show_bug.cgi?id=661286
https://git.gnome.org/browse/mutter/commit/?id=28deea4
https://git.gnome.org/browse/metacity/commit/?id=c9099b4
https://github.com/mate-desktop/marco/issues/205


NiceandGently:
confirmed this fixes the issue

PR #211, opened 2015-08-01
window menu: make minimize item insensitive when window should not be minimized

monsta:
(e.g. delete confirmation dialog, logout dialog...)

it was working that way in gnome 2, I don't know why perberos commented it out...

fixes https://github.com/mate-desktop/caja/issues/443

@NiceandGently @infirit @flexiondotorg
please test it for any possible regressions :smile:


NiceandGently:
Does what it promised, i see no **collateral damage** after 20min running marco :)

infirit:
It may not have been perberos but unfortunately a lot of history got lost so we can't check.

PR #219, opened 2015-09-14
Add manual and by-pointer window placement

### LINKED ISSUE #123: Pointer window placement
xfce has a feature where windows are being placed according to the pointer position at the time. It's incredibly nifty and useful, at least to me :)

muffin, a mutter fork, has implemented this as well, with
https://github.com/linuxmint/muffin/commit/325767132d22a008a7b9c397d9ca1751bac41a62

The code looks fairly similar to marco (due to muffin/mutter/metacity common inheritance) and the patch is fairly trivial (basically one function + stitching that to the code with an option).

It could probably very easily be ported to marco and I could give it a stab, although someone with more marco experience would probably tackle it in far less time than I would :)

paravoid:
This adds a window placement preference: the existing behavior is now
called "automatic" and is the default. Two new modes are being
introduced: "pointer", which means that windows are placed according to
the mouse pointer position; and "manual" which means that the user must
manually place the new window with the mouse or keyboard.

This is a straight port from muffin, see commit https://github.com/linuxmint/muffin/commit/325767132d22a008a7b9c397d9ca1751bac41a62
Fixes #123.

### DIFF
--- a/src/core/place.c
+++ b/src/core/place.c
@@ -87,6 +87,47 @@ northwestcmp (gconstpointer a, gconstpointer b)
     return 0;
 }
 
+static gboolean
+place_by_pointer(MetaWindow *window,
+                 MetaFrameGeometry *fgeom,
+                 MetaPlacementMode placement_mode,
+                 int *new_x,
+                 int *new_y)
+{
+  int window_width, window_height;
+  Window root_return, child_return;
+  int root_x_return, root_y_return;
+  int win_x_return, win_y_return;
+  unsigned int mask_return;
+
+  XQueryPointer (window->display->xdisplay,
+                 window->screen->xroot,
+                 &root_return,
+                 &child_return,
+                 &root_x_return,
+                 &root_y_return,
+                 &win_x_return,
+                 &win_y_return,
+                 &mask_return);
+
+  window_width = window->frame ? window->frame->rect.width : window->rect.width;
+  window_height = window->frame ? window->frame->rect.height : window->rect.height;
+
+  if (fgeom) {
+    *new_x = root_x_return + fgeom->left_width - window_width / 2;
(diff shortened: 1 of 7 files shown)


paravoid:
Also note that this is a feature that xfwm4 has so it's not completely unheard of outside of muffin.

I also have a port of the "center" window placement as a follow-up to this commit, see
 https://github.com/paravoid/marco/tree/window-placement-center
I've never used the "center" placement though and I can't say that I've verified it by trying it so take it with a grain of salt (I can submit a separate PR).

infirit:
Works nicely. The only thing is the description in the schema should be one long string, not broken up with newlines.

If you are up for it maybe add this to the "Placement" tab in the "New Window" section in the [window preferences](https://github.com/mate-desktop/mate-control-center/blob/master/capplets/windows/mate-window-properties.c) capplet?

paravoid:
Updated. I'll have a look at the window preferences capplet next.

infirit:
Thank. I have this commit locally in a branch updated for other changes I have in https://github.com/mate-desktop/marco/pull/220. As soon as that is pushed to master I'll push this one as well.

paravoid:
Thanks.

Over there I see a migration from MetaFrameGeometry to MetaFrameBorders; did you actually this commit for itas well?

infirit:
I pushed it to my fork so you can have a look, https://github.com/infirit/marco/commit/74a31e3f1a6e34b91d5b86e82432b590aed2ee7c. Build tested only but should be ok :smile:.

paravoid:
Great, thanks :)

infirit:
I decided to merge now and update my pull request. Merged as https://github.com/mate-desktop/marco/commit/dee034771c8f3b519c8011faf78d277d6af76933.

Thanks

PR #223, opened 2015-09-18
gtk3: draw tile border

dnk:
draw non transparent tile border.
With border preview window looks more completely.


raveit65:
Works great, thank you.
I looks much better with a thiny border if compositor is on.
To have a StylingWidget for styling it in themes is a dream :)
@flexiondotorg
please test and merge

Issue #200, opened 2015-06-13
Marco 1.10.1 crash with SIGABRT


Th30n:
This started happening as I upgraded to 1.10 and is still happening in 1.10.1.
I'm using ArchLinux, and crashes so far happenened only when switching windows from/to Steam windows (so it might be related). Also the xorg-server was upgraded to 1.17.1, and gcc is 5.1.
Compositing is turned on, perhaps this is related to issue #181.
Unfortunately I don't have debug symbols, but here is the backtrace:

    Core was generated by `marco'.
    Program terminated with signal SIGABRT, Aborted.
    #0  0x00007fd05e7c9528 in raise () from /usr/lib/libc.so.6
    (gdb) bt
    #0  0x00007fd05e7c9528 in raise () from /usr/lib/libc.so.6
    #1  0x00007fd05e7ca93a in abort () from /usr/lib/libc.so.6
    #2  0x00000000004374f8 in ?? ()
    #3  0x00000000004246d4 in ?? ()
    #4  0x00007fd05f4b287d in _XError () from /usr/lib/libX11.so.6
    #5  0x00007fd05f4af7c7 in ?? () from /usr/lib/libX11.so.6
    #6  0x00007fd05f4af885 in ?? () from /usr/lib/libX11.so.6
    #7  0x00007fd05f4b0830 in _XReply () from /usr/lib/libX11.so.6
    #8  0x00007fd05f495c35 in XGetImage () from /usr/lib/libX11.so.6
    #9  0x0000000000423d18 in ?? ()
    #10 0x0000000000423fce in ?? ()
    #11 0x00007fd05fe2a3a3 in ?? () from /usr/lib/libglib-2.0.so.0
    #12 0x00007fd05fe2990d in g_main_context_dispatch () from /usr/lib/libglib-2.0.so.0
    #13 0x00007fd05fe29ce0 in ?? () from /usr/lib/libglib-2.0.so.0
    #14 0x00007fd05fe2a002 in g_main_loop_run () from /usr/lib/libglib-2.0.so.0
    #15 0x000000000040f627 in ?? ()
    #16 0x00007fd05e7b6790 in __libc_start_main () from /usr/lib/libc.so.6
    #17 0x000000000040f9d9 in ?? ()

NiceandGently:
@Th30n 
can you please run gdb with 'thread apply all bt full', see
http://fedoraproject.org/wiki/StackTraces#gdb

Th30n:
[...]

Issue #228, opened 2015-10-22
marco crashes when moving a window with a mouse


raveit65:
Here a litle video which describes the issue:
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/marco-crash.ogv
Steps to reproduce:
1. enable compositor
2. start a new mate-session
3. open min 2 windows
4. move one window
5 boom, marco crashed

Note, this doesn't happens with gtk3 build and if compositor is off in gtk2 build.
It's reported in all fedora versions, f21/f22/f23
Downstream report:
https://bugzilla.redhat.com/show_bug.cgi?id=1258131 (other reports are marked as duplicates)
Stacktrace fom today:
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/marco-backtrace
```
Truncated backtrace:
Thread no. 1 (10 frames)
 #4 split_edge at core/boxes.c:1569
 #5 meta_rectangle_remove_intersections_with_boxes_from_edges at core/boxes.c:1691
 #6 compute_resistance_and_snapping_edges at core/edge-resistance.c:1089
 #7 apply_edge_resistance_to_each_side at core/edge-resistance.c:556
 #8 meta_window_edge_resistance_for_move at core/edge-resistance.c:1172
 #9 update_move at core/window.c:7182
 #10 meta_window_handle_mouse_grab_op_event at core/window.c:7654
 #11 event_callback at core/display.c:2084
 #12 filter_func at ui/ui.c:289
 #13 gdk_event_apply_filters at gdkevents-x11.c:371
```

flexiondotorg:
This is reproducible in Ubuntu MATE 15.10. Also crashed in Animations are enabled when Maximising/Minimising Google Chrome or Chromium, although in this case compositing can be on or off.

raveit65:
yes, in some reports reporter talks about resizing too.

raveit65:
We have a regression introduced somewere between 1.9.2 and 1.9.3.
marco-1.9.2-2.fc22 works well.
marco-1.9.3-1.fc22 crashed.
I will track down it more with git bisect.

raveit65:
Ok, the bad commit is http://git.mate-desktop.org/marco/commit/?h=1.10&id=bcbc109
**add support for _GTK_FRAME_EXTENTS**

raveit65:
reverting those 3 commits via git rebase works cleanly and fixes the issue for me.
[...]

PR #229, opened 2015-10-22
effects: remove hack that made Marco crash in certain situations

monsta:
fixes https://github.com/mate-desktop/marco/issues/200

ported to Marco from https://github.com/GNOME/metacity/commit/f25b7760b7430b674afdb4de1e2072cd99ea3fbc

### DIFF
--- a/src/core/effects.c
+++ b/src/core/effects.c
@@ -316,23 +316,6 @@ update_wireframe_window (MetaDisplay         *display,
 }
 #endif
 
-/**
- * A hack to force the X server to synchronize with the
- * graphics hardware.
- */
-static void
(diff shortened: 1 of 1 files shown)


raveit65:
Seems to be fine, no regression found after testing a while.
But i was never effected by resizing issue, so i can't confirm if this is fixed.
anyways, looks good.
+1

Btw. it doesn't fixes https://github.com/mate-desktop/marco/issues/228

monsta:
Oh but https://github.com/mate-desktop/marco/issues/228 is a separate bug, I can't magically fix all the bugs during one day :smile:

Issue #208, opened 2015-07-26
GTK3: Some metacity themes are not rendered correctly


City-busz:
If I select e.g. the Shiny theme, the window borders and the inactive title bars are black instead of the appropriate color.

infirit:
This is a theme issue plus shiny was pretty broken in Gtk3 and removed from mate-themes.

City-busz:
Are you sure? The latest Metacity with GTK3 is able to render the shiny theme correctly, so I think it's a bug in Marco.

NiceandGently:
The question is what does shiny metacity theme read?
With a complete theme the color definitions in metacity.xml , ie, color=gtk:fg[NORMAL], are read from gtkrc in case of gtk2, or from settings.ini in case of gtk3.
So, shiny doesn't have this files anymore.
Btw, latest metacity-3.6.1 use server-side-decorations, means all applications don't use the metacity theme anymore and looks like csd-applications.

City-busz:
Marco tries to read the colors from the currently selected GTK+ theme:
https://github.com/mate-desktop/marco/blob/master/src/ui/theme.c#L1490

But this is broken in some cases, and wrong colors are used.

NiceandGently:
I think it is valid to use fix colors in metacity.xml, a lot of themes do that.
But i to revert those commits if the issue is fixed.
By the way, metacity doesn't displays the color of submarine metacity theme correct, if using color variables.

infirit:
Does https://github.com/infirit/marco/commit/9ffd9fdf2a256387ec428a5b094d546b1cb653f2 fix the problem?

NiceandGently:
This helps to fix bg color of unfocus windows, but font colors and colors for buttons are ignored.
![marco-t-o](https://cloud.githubusercontent.com/assets/961604/8895063/b4cbc96a-33cc-11e5-9dd8-b14f0e3163e0.png)

infirit:
Interesting, we should be in the ```META_GTK_COLOR_FG``` part of ```meta_set_color_from_style``` which calls ```gtk_style_context_get_color (context, state, color);```. When I run this in python I get black (#000000), which is correct. But for some reason we get white in Marco. tbc

```
>>> context.get_color(Gtk.StateFlags.NORMAL).to_string()
'rgb(0,0,0)'
```

monsta:
[...]

PR #240, opened 2015-11-17
Gtk3: fix colors

City-busz:
These changes are required to completely fix issue #208 in addition to commit 6ca27a47.

Gtk3: Fix foreground color
Similar to the previous background color fix, we need to fix text color too.

Gtk3: Always read colors
Without reading GTK+ colors every time, the colors remain unchanged, when the user changes the actual GTK+ theme.


City-busz:
I found a much better one-line fix for the same issue. See: #241.

PR #252, opened 2016-02-05
Bump libmarco-private's sover in Gtk3 build

XRevan86:
Differences between Gtk2 and Gtk3 builds of libmarco-private are even bigger than linked GTK+ version, libmarco-private's API itself is different.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -125,7 +125,11 @@ libmarco_private_la_SOURCES = \
 	ui/theme.c \
 	ui/theme.h
 
-libmarco_private_la_LDFLAGS = -no-undefined
+if HAVE_GTK3
+libmarco_private_la_LDFLAGS = -no-undefined -version-info 1:0:0
+else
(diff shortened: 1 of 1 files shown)


monsta:
Maybe bump both versions, using 2 for GTK+2 build and 3 for GTK+3 build? It's something that gnome devs already do I think...

XRevan86:
@monsta, there's no need to bump without an ABI breakage. So "0" for Gtk2 build is fine.

monsta:
No problem, I just thought it would be more descriptive :smile:

XRevan86:
… until a second API bump :-).

raveit65:
Thank you

monsta:
Hmm, this requires a rebuild of mate-control-center in GTK+3 (or else mate-window-properties will try to load ```libmarco-private.so.0``` instead of ```libmarco-private.so.1```).

raveit65:
Well, it's currently only in master, we can increase marco required version there.
And the same if we want to packport it to 1.12.

PR #255, opened 2016-02-28
GTK+3: Get values from GtkStyleContext states properly

raveit65:
this avoids warnings like this:
Warning of Windowmanagement:Log level 16: State 4 for context 0x55988f9ad950
doesn't match state 128 set via gtk_style_context_set_state ()

### DIFF
--- a/src/ui/tile-preview.c
+++ b/src/ui/tile-preview.c
@@ -143,8 +143,13 @@ on_preview_window_style_set (GtkWidget *widget,
   gtk_widget_path_append_type (path, GTK_TYPE_ICON_VIEW);
   gtk_style_context_set_path (context, path);
 
-  gtk_style_context_get (context, GTK_STATE_FLAG_SELECTED, "background-color", &preview->preview_color, NULL);
+  gtk_style_context_save (context);
+  gtk_style_context_set_state (context, GTK_STATE_FLAG_SELECTED);
+  gtk_style_context_get (context, gtk_style_context_get_state (context),
+                         "background-color",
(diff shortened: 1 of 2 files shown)


raveit65:
@ monsta
@XRevan86 
Can you please test this as marco is a base application of Mate?

monsta:
Wow this is cumbersome... save context, set state, get some value, restore context... :open_mouth: 
Is it required in GTK+3 these days?

@lukefromdc please check this too.

raveit65:
> Is it required in GTK+3 these days?
Yes, with gtk+-3.19 you see those warnings with a bunch of applications.
Btw. all is taken from :) https://github.com/raveit65/compiz/commit/3440ee033fa2691fd90c1e5d8eebf328062cf63a

lukefromdc:
I built and tested this, no obvious problems. Didn't see any obvious WM manager warnings
running apps from terminal either way, but did get fewer warnings on running marco --replace,
getting only this one:

Window manager warning: Log level 16: g_object_set_valist: object class 'MetaFrames' has no property named 'type'

monsta:
No problems with GTK+ 3.18.8. That said, I've never seen these warnings about ```gtk_style_context_set_state``` - maybe they're turned on only in dev versions like 3.19...

raveit65:
This simply new with gtk+-3.20.
And those warning say clear that something is wrong and needs to improved :smile: 
Again, the link to M.Classen blog
https://blogs.gnome.org/mclasen/2015/11/20/a-gtk-update/

Issue #232, opened 2015-11-11
GTK3: Alt-Tab switcher does not shows previews.


pyaazkaparatha:
Alt Tab window switcher shows only icons no previews. Enabling and disabling thumbnails in window preferences does not help. 

marco-1.12.0-12.fc23.x86_64 in Fedora 23 with GTK3 repos enabled.

raveit65:
confirmed,
as a note, with gtk2 build the tab-preview works

PR #256, opened 2016-03-03
[GTK+3] re-add ability to show alt-tab thumbnails when compositor is on

monsta:
fixes https://github.com/mate-desktop/marco/issues/232

@raveit65 @lukefromdc
please test this big change :smile:

### DIFF
--- a/src/compositor/compositor-private.h
+++ b/src/compositor/compositor-private.h
@@ -44,8 +44,13 @@ struct _MetaCompositor
   void (*process_event) (MetaCompositor *compositor,
                          XEvent         *event,
                          MetaWindow     *window);
+#if GTK_CHECK_VERSION (3, 0, 0)
+  cairo_surface_t *(* get_window_surface) (MetaCompositor *compositor,
+                                           MetaWindow     *window);
+#else
   Pixmap (*get_window_pixmap) (MetaCompositor *compositor,
                                MetaWindow     *window);
+#endif
   void (*set_active_window) (MetaCompositor *compositor,
                              MetaScreen     *screen,
                              MetaWindow     *window);
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -34,6 +34,11 @@
 #include <unistd.h>
 
 #include <gdk/gdk.h>
+#include <gtk/gtk.h>
+
+#if GTK_CHECK_VERSION (3, 0, 0)
+#include <cairo/cairo-xlib.h>
+#endif
 
 #include "display.h"
 #include "screen.h"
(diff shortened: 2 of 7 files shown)


raveit65:
Thumbnails are working here again, good job :)
But i'm a bit confused with the gsettings keys. compositing-fast-alt-tab enables the setting, right?
I thought there was a other key for it.
And the compositor effect is always there, with or w/o compositing-fast-alt-tab enabled.
:confused:

raveit65:
I was wrong, compositing-fast-alt-tab disables thumbnail preview, but  the compositor effect is always there.

lukefromdc:
I'll test the changes-but alt-tab is something I have never used, so all I
can really do is ensure that nothing breaks in gtk3.19

lukefromdc:
Just built and tested this, everything seems to work OK on GTK 3.19. With composition on (my default) hitting alt-tab will show window thumbnails and a clear outline of one window, whose contents are displayed when alt-tab is release. 

I do get this warning on startup from marco --replace in terminal:

`Window manager warning: Log level 16: g_object_set_valist: object class 'MetaFrames' has no property named 'type'`

monsta:
@lukefromdc: I get that warning as well with 3.18, it was there before my changes...

@raveit65: yes, that's a key for toggling thumbnails only. The key for toggling compositor itself is ```compositing-manager```.

raveit65:
Hmm, weird, i don't have a key for toogling the thumbnails
```
[rave@mother ~]$ gsettings list-recursively org.mate.Marco.general
org.mate.Marco.general
[rave@mother ~]$ gsettings list-recursively org.mate.Marco.general
org.mate.Marco.general application-based false
org.mate.Marco.general disable-workarounds false
org.mate.Marco.general action-right-click-titlebar 'menu'
org.mate.Marco.general mouse-button-modifier '<Alt>'
org.mate.Marco.general focus-new-windows 'smart'
org.mate.Marco.general wrap-style 'no wrap'
org.mate.Marco.general action-middle-click-titlebar 'lower'
org.mate.Marco.general reduced-resources false
org.mate.Marco.general theme 'Blue-Submarine'
org.mate.Marco.general num-workspaces 8
org.mate.Marco.general raise-on-click true
org.mate.Marco.general visual-bell-type 'fullscreen'
org.mate.Marco.general auto-raise true
org.mate.Marco.general visual-bell false
org.mate.Marco.general side-by-side-tiling false
org.mate.Marco.general focus-mode 'sloppy'
org.mate.Marco.general action-double-click-titlebar 'toggle_maximize'
org.mate.Marco.general compositing-manager true
org.mate.Marco.general titlebar-uses-system-font false
org.mate.Marco.general auto-raise-delay 500
org.mate.Marco.general button-layout 'menu:minimize,maximize,close'
org.mate.Marco.general compositing-fast-alt-tab false
org.mate.Marco.general placement-mode 'automatic'
org.mate.Marco.general audible-bell true
org.mate.Marco.general center-new-windows false
org.mate.Marco.general titlebar-font 'Sans Bold 10'
org.mate.Marco.general resize-with-right-button true
```
and as i said enable 'compositing-fast-alt-tab' breaks thumbnailing and 
it looks like that the ugly effect of 'compositing-fast-alt-tab' is always on.
Please test again.

monsta:
Wait, but it's ```compositing-fast-alt-tab``` key that's used to toggle thumbnails displaying... it means that icons are displayed instead of thumbs. Maybe the name isn't good (metacity has different name), but it's indeed faster as no additional work is done by Marco.

raveit65:
I think 'compositing-fast-alt-tab' should enable only the ugly compositior effect, which is always on.
And enable this key disabled thumbnailing.

monsta:
You mean shadowing effect? Hmm... the code is taken from metacity, so the effect should be the same there... Also @lukefromdc didn't mention that something is wrong with the effect.

raveit65:
> You mean shadowing effect?

Of course not :smile: 
In f23 with marco-1.12.1-3.fc23.x86_64 (gtk2)
Thumbnailing is enabled with enable the compositor, *'compositing-manager'* key.
I see no effect on windows if *compositing-fast-alt-tab* is disabled, only a frame which marks the selected window.
Enable this key disable *thumnailing* but no effect on window.
So, it looks like that:
1. the effect on windows is different between both toolkit versions, or this is broken in gtk2.
2. The key *compositing-fast-alt-tab* means **thumbnailing** but here the enable/disable switch is reversed.

PR #261, opened 2016-04-07
Integrate upstream patches from metacity for new-window-always-on-top

slseed1969:
These changes correspond with the fixes associated with Red Hat Bugzilla #530263

The normal behavior is that if a new window is not given the
  focus (since, for example, the user has interacted with another
  window after launching an application), then if the window would
  be stacked on top of the focus window, the window is instead
  stacked beneath and flashed in the taskbar. This behavior can
  be annoying on large screens and multihead setups where the
  taskbar is hard to notice and difficult to get to, so this option,
  if set, disables this behavior, and new windows are always placed
  on top, whether or not they get focus.

  Note that if this option is set, a new window may completely hide
  the focus window but not get focus itself, which can be quite confusing
  to users. Also, note that setting this option breaks the normal
  invariant in the 'click' focus mode that the topmost window always
  has focus, so its most suitable for use with the 'mouse' and
  'sloppy' focus modes.

Upstream bug is https://bugzilla.gnome.org/show_bug.cgi?id=599261

### DIFF
--- a/src/core/prefs.c
+++ b/src/core/prefs.c
@@ -95,6 +95,7 @@ static MetaVirtualModifier mouse_button_mods = Mod1Mask;
 static MetaFocusMode focus_mode = META_FOCUS_MODE_CLICK;
 static MetaFocusNewWindows focus_new_windows = META_FOCUS_NEW_WINDOWS_SMART;
 static gboolean raise_on_click = TRUE;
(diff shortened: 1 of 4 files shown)


raveit65:
This breaks  expectation on click focus, was never merged in metacity and it has conflicts with actual master.
Thanks for reporting but we want to close it.

PR #308, opened 2017-01-20
Added hotkey to rename current workspace

fbarriga:
Add support for a hotkey that can rename the current workspace.

### DIFF
--- a/src/50-marco-desktop-key.xml.in
+++ b/src/50-marco-desktop-key.xml.in
@@ -11,5 +11,7 @@
 
 	<KeyListEntry name="run-command-terminal" _description="Run a terminal" />
 
+	<KeyListEntry name="rename-workspace" _description="Rename Current Workspace" />
+
 </KeyListEntries>
 
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -38,6 +38,8 @@
 #include "effects.h"
 #include "util.h"
 
+#include <gio/gio.h>
+
 #include <X11/keysym.h>
 #include <string.h>
 #include <stdio.h>
@@ -64,6 +66,8 @@ handler (MetaDisplay    *display,\
          XEvent         *event,\
          MetaKeyBinding *binding);
 #include "all-keybindings.h"
+#include "../include/util.h"
+
 #undef keybind
 
 /* These can't be bound to anything, but they are used to handle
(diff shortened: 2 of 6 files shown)


raveit65:
Is there a reason why you bump version again in your PRs ?

fbarriga:
@raveit65 bump version removed and branch rebased.

monsta:
This looks quite complicated. Adding lots of code to launch zenity and handle its output... does this really belong to Marco?

raveit65:
Any progress here?

raveit65:
@fbarriga 
I like to see that feature in marco, but we have an open question here from a core-team member here?

monsta:
I think it will be safe to close it for now.

raveit65:
no, please do not, i like that feature!!

monsta:
Ok, I just suggested :slightly_smiling_face:

fbarriga:
@monsta I didn't like either have to use `zenity` for this but I didn't found any other solution. I decided to use it because it was the solution being used in function `meta_show_dialog`.

raveit65:
Ok, that's declares it, i will  merge it the next days if nobody gives his veto ;-)
It works for me and i see no collateral damage.

monsta:
Yes, zenity is used for showing a simple message dialog outside of marco's process, but there's no feedback like here (when some data is transferred from the dialog back to marco). I was wondering if it could possibly block marco's process or affect it in some other way.

monsta:
For now it seems to work here. Maybe @lukefromdc could give it a try as well, just in case.

lukefromdc:
I have never named my workspaces, to test this I would need to know how to undo that

monsta:
Oh, that's easy: they're in `org.mate.Marco.workspace-names` schema, so you can just revert the keys to the default values if needed.

raveit65:
@lukefromdc 
After renaming workspace 2 i get those gsettings value.
```
$ gsettings get org.mate.Marco.workspace-names name-2
'Sun of a .....'
```
With
```
$ gsettings reset org.mate.Marco.workspace-names name-2
```
i restore the default.

raveit65:
opps, someone was faster :-)

raveit65:
merged https://github.com/mate-desktop/marco/commit/5ded52f5e98d60abb903cc9194714a9b293ad482
Thank you.

monsta:
No need to edit the schema, this new keybinding shows up in mate-keybinding-properties. :slightly_smiling_face: 

That said, maybe we should move it from Desktop to Window Management category? All workspace-related bindings are there.

PR #312, opened 2017-02-07
Dev deprecations

raveit65:


raveit65:
Note,  theme-viewer: port GtkAction to GAction is partial from https://git.gnome.org/browse/metacity/commit/?id=7012fd2b5a166ac552aa715601d7a4cf6fb85b3b

monsta:
Hmm, all sample "windows" are fully expanded horizontally now. Previously GtkAlignment made them expand half the size of the main window, due to ```xscale``` and ```yscale``` arguments set to 0.5 (see their meaning [here](https://developer.gnome.org/gtk3/stable/GtkAlignment.html#gtk-alignment-new)).

Before:
marco-before

After:
marco-after

Wonder if the same effect can be achieved via some modern API... For this app it's ok, it's just a demo. I think you can merge it. But we should remember about this for other apps.

raveit65:
I will try to uses GTK_ALIGN_CENTER instead of GTK_ALIGN_FILL, see https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkAlign
And playing bit with the gtk_box_pack_start options seems useful.
https://developer.gnome.org/gtk3/stable/GtkBox.html#gtk-box-pack-start

raveit65:
Using GTK_ALIGN_CENTER moves the window to the middle and it is to small.
Well, i will merge it like it is.

monsta:
I know, I already tried to play with widget's alignment and expand options, but couldn't get the same result as with GtkAlignment's options.

PR #327, opened 2017-06-20
Fix verbose bug + restore mate-panel shadows

ikeydoherty:
Title says it all, really.

screenshot from 2017-06-20 22-03-29

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -45,6 +45,7 @@
 #include "window.h"
 #include "compositor-private.h"
 #include "compositor-xrender.h"
+#include "../core/window-private.h"
 #include "xprops.h"
 #include <X11/Xatom.h>
 #include <X11/extensions/shape.h>
@@ -883,6 +884,14 @@ window_has_shadow (MetaCompWindow *cw)
         return TRUE;
       }
     }
+
+  /* Calculate mate-panel *prior* to ARGB checks */
+  if (cw->type == META_COMP_WINDOW_DOCK && cw->window) {
+    if (cw->window->res_name && g_str_equal(cw->window->res_name, "mate-panel")) {
+      meta_verbose ("Window has shadow because it is mate-panel");
+      return TRUE;
+    }
+  }
 
   /* Do not add shadows to ARGB windows */
   if (cw->mode == WINDOW_ARGB) {
--- a/src/ui/frames.c
+++ b/src/ui/frames.c
@@ -118,10 +118,6 @@ meta_frames_constructor (GType                  gtype,
   gobject_class = G_OBJECT_CLASS (meta_frames_parent_class);
   object = gobject_class->constructor (gtype, n_properties, properties);
 
-  g_object_set (object,
-                "type", GTK_WINDOW_POPUP,
-                NULL);
-
   return object;
 }
 


ikeydoherty:
Please note the commit description in d281b76fb24ea3e39bba8966218611575fa1a31b as it's highly important imho. mate-panel will also need some magic added to it so that we avoid shadows on pure alpha cases (not theme based alpha)

ikeydoherty:
As an aside it might be an improvement to only use the `SMALL` shadow type for a top dock and `MEDIUM` elsewhere, due to the offset + direction of the shadow

ikeydoherty:
OK so it turns out we can only really do that logic at startup in mate-panel because we cannot unrealize() the window at all (we'll murder all the applets and upset them deeply)

However I have a PR to apply this at startup in mate-panel which I'll file separately

ikeydoherty:
No, I'm not wrong.. Please look at MetaFrames API and notice it has no _set_property/get_property, and is a subclass of GtkInvisible. There is *no* `type` parameter, as can be seen when gobject craps its pants in the terminal for using the wrong parameter.

As for "why not use extra / new atom" - because this is restoring behaviour that already existed, not adding *new* functionality. Effectively fixing a regression.

muktupavels:
Oh, did not know they use GtkInvisible. Sorry!

But is WM_CLASS right thing to use to fix regression? From GTK+ doc - "According to the ICCCM, you should always set these to the same value for all windows in an application". If user use different styles for top and bottom panels then these values will not be same!

ikeydoherty:
The class matching is only applicable to windows that don't set the skip hints, this is a window that skips taskbar + pager, sets `_NET_WM_WINDOW_TYPE_DOCK`, and would not be grouped in any kind of Application Leader/Window flow.

raveit65:
Can you provide some simple steps to test this commit?
Before and after?

raveit65:
Cool, first commit fixes this annoying warning which spams  .xsession-errors since years ;-)
```
[rave@mother ~]$ marco --replace

** (marco:13468): WARNING **: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-cMlGxEZ9Am: Connection refused
Warnung der Fensterverwaltung:Log level 16: g_object_set_valist: object class 'MetaFrames' has no property named 'type'
```

ikeydoherty:
So the first one is definitely a take-it no-bother kinda of no brainer. The second one I'd like to discuss because there is still *a* bug, and this might be a slightly immoral use of wmclass. Long story short if Marco starts *after* mate-panel, it'll identify it and see it as `mate-panel` wmclass.

The alternative is that instead, we dynamically adjust the GdkVisual used for mate-panel and drop ARGB from it, but that prevents semi transparent themes. There is still this race condition on the startup though that you must have the panel open before the WM, which has also persisted for years (Maybe I'll dig in this during the weekend?)

It has become obvious from social media that some people still want to be able to have no shadow though, hence the dynamic approach

muktupavels:
Just introduce new atom / property for this. Should not be hard to set / update that from mate-panel. Then update compositor to read this value. Something similar to get_window_type.

raveit65:
Ok, second commit sets the shadow to the panel if marco starts after mate-panel, but panel transparency from preferences is complete broken.
Restarting the panel at this point brings back the transparency of the panel, but shadow is gone.
Now, i will add mate-panel PR to this test setup.

ikeydoherty:
You have to restart mate-panel to change the wmclass which is why the wmclass stuff is non ideal

raveit65:
Hmm, after applying m-p PR on top of 1.19.1 release i see see no different.
If marco starts before m-p i have a shadow but transparency is broken.
If m-p starts first (= restart mate-panel) than there is no shadow and transparency is fine.
I also use this PR on top of 1.19.1 release https://github.com/mate-desktop/mate-panel/pull/597
Not sure if this is what you want.

raveit65:
New day --> new session start --> i don't see the shadow except i restart marco if m-p is running.
And transparency isn't usable.
Maybe your results are different to mine, but panel transparency is a very amazing feature since we use gtk+3 and i don't want to loose it.
@monsta @lukefromdc 
thoughts?

raveit65:
I think i will cherry-pick first commit to master if all agree.

muktupavels:
How it breaks transparency? Is there black color under panel? Can you post screenshot?

lukefromdc:
I absolutely would NOT want to see a shadow forced on the panel unless themes can override it by setting box shadow to none or to zero values. In my earlier test of the previous pair of commits I did not see any panel shadows, but maybe no theme I had made them, mine certainly does not.

raveit65:
panel with shadow using normal background from theme.  👍 
mate-panel-with-shadow
panel with shadow using transparency from preferences. 👎 
mate-panel-with-shadow-and-transparency

This is with Menta themes, our default.

raveit65:
For compare, a transparency panle should look like this.
mate-panel-transparency

lukefromdc:
That's an ugly mess, and appears to be a shadow being rendered underneath the panel.

muktupavels:
It is shadow; compositor creates shadow picture for whole window! When drawing shadow you must exclude window area so only "visible" part (outside of real window) is drawn.

raveit65:
> In my earlier test of the previous pair of commits I did not see any panel shadows

Try to restart marco after m-p is already running.

raveit65:
> I absolutely would NOT want to see a shadow forced on the panel unless themes can override it by setting box shadow to none or to zero values.

I full agree with you, but for me it's sounds better to add a setting for enable the shadow in preferences which is independent from the theme?

lukefromdc:
That would work too. 

Still no luck getting a shadow to show up here though-and just to test it I have to rearrange all my open windows to one desktop, switch from compiz to marco, then set everything back afterwards.

raveit65:
I cherry-picked first commit 86153933369f9fc2896542e567c347fa886d134c

ikeydoherty:
OK so it seems the general consensus then that the shadow becomes a separate setting (i.e. gsetting), and we'd implement some kind of new atom to request this per @muktupavels's suggestion. Also while doing that we'd want to track down the race condition, so that regardless of who starts first, the shadowing logic is legit.

muktupavels:
I think there will be no race condition! If panel is started before marco then marco will read atom that is set by panel. If that atom is not available by time when panel is known to compositor then it will get PropertyNotify event.

There is also bug! If panel changes its window type to DOCK after it is known to compositor it will not be added to dock_windows lists.

To remove shadow under panel something like this should work in paint_dock_shadows:
```
XserverRegion region;
XRectangle r;

r.x = panel x pos;
r.y = panel y pos;
r.width = panel width;
r.height = panel height;

region = XFixesCreateRegion (xdisplay, &r, 1);
XFixesSubtractRegion (xdisplay, shadow_clip, shadow_clip, region);
```

ikeydoherty:
> There is also bug! If panel changes its window type to DOCK after it is known to compositor it will not be added to dock_windows lists.

Like I said. Race condition. :) In reality the window should be constructed with the dock hint.

raveit65:
Honestly, i am not sure what is starting first.
In theory the 'phase WINDOW_MANAGER' is before 'phase PANEL' in mate-session logs.
But parts of marco entries are in  'phase PANEL' here.
You can read session logs in xsession-errors if you enable debug.
```
[rave@f26 ~]$ gsettings get org.mate.debug mate-session
true
```

ikeydoherty:
`marco --replace&`

raveit65:
Yes, this command restarts marco, but this isn't new ;-)

ikeydoherty:
I didn't say it was new, I was telling you to restart it to debug it.

raveit65:
I think you complete miss understood my previous comment, but it's OK, no need to spam this report here.

mrnhmath:
What's the consensus on this?

PR #328, opened 2017-06-22
Fix setting name typo

jamesward:
The setting name was incorrect:
```
jw@t470p:~$ gsettings list-recursively |grep num-work
org.gnome.desktop.wm.preferences num-workspaces 4
org.mate.Marco.general num-workspaces 1
org.mate.Marco.general num-workspaces 1
```


monsta:
Thanks for spotting it. There were more similar mistakes in GSettings keys names, so I corrected them all.

Issue #182, opened 2015-03-14
Add ability to use the Arrow Keys when switching windows with ALT+TAB [$100]


BrootLee:
Hello.
Take any popular desktop environment you like. You can use the Arrow Keys when switching windows with ALT+TAB. It's more comfortable and faster than having to press TAB several times with the same finger. Please add this feature.
Thank you.
##

imperative:
Additionally, it should be possible to use mouse to select the application to switch to, within the alt-tab window. (By clicking on the icon of that application.). This is also the default and very useful and userfriendly behavior, which exists in almost any other window manager.

BrootLee:
I'm so glad someone else became interested in this! I agree with the above comment by the way, using the mouse to select the window should also be possible.

PapayaMedia:
I don't remember to have used the arrow keys to move between window thumbnails. But the mouse option to select the needed window is a must have. I really want this feature to be implemented.

yoburtu:
Hello,

Has any of this request been incorporated into Mate Desltop 1.18? I am very interested in it.

Best regards.

raveit65:
merged
https://github.com/mate-desktop/marco/commit/c4750c85961777e6ab00b46fb12b1ba0253796d5
https://github.com/mate-desktop/marco/commit/7c69bcf4d5b195902c770bcbf84bd8b8c3427059

imperative:
Does the mouse work now too? Or should we create another issue?

PR #330, opened 2017-06-24
Allow use of arrows while in alt+tab popup

pldubouilh:
(issue #182)

### DIFF
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -2220,9 +2220,9 @@ process_tab_grab (MetaDisplay *display,
     }
 
   /* Allow use of arrows while in window switching mode */
-  if (event->xkey.keycode == 114)
+  if (event->xkey.keycode == ARROW_RIGHT || event->xkey.keycode == ARROW_RIGHT_PAD)
     key_used = TRUE;
-  else if (event->xkey.keycode == 113)
+  else if (event->xkey.keycode == ARROW_LEFT || event->xkey.keycode == ARROW_LEFT_PAD)
     {
       key_used = TRUE;
       backward = TRUE;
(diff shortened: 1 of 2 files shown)


raveit65:
@pldubouilh 
We have a open question from core-team member.

pldubouilh:
@monsta @raveit65 I added support for the numpad arrows too, but I don't have such a keyboard at hand so please test it

raveit65:
I just tested it. Arrow keys on numpad are usable now.
Thank you

monsta:
Works fine, but I accidentally noticed a funny thing...
You know that holding Shift in addition to Alt makes Tab switch the windows in reverse order. Guess what, now holding Shift swaps left and arrows too. 🙂 
Are you guys ok with that?

raveit65:
The function of holding shift is to reverse something, so from logic point of view this is fine.
And i don't think that someone want to use the arrow keys and shift key at the same time.
We shouldn't make it more complicated with disabling the shift key if the arrow keys are used ;-)
I am OK with current state of PR.

monsta:
> The function of holding shift is to reverse something, so from logic point of view this is fine.

I know, it's just funny how it affects arrow keys :slightly_smiling_face:

raveit65:
merged
https://github.com/mate-desktop/marco/commit/c4750c85961777e6ab00b46fb12b1ba0253796d5
https://github.com/mate-desktop/marco/commit/7c69bcf4d5b195902c770bcbf84bd8b8c3427059
Thank you

raveit65:
@pldubouilh 
Don't forgot to claim the bounty here.
https://www.bountysource.com/issues/9539963-add-ability-to-use-the-arrow-keys-when-switching-windows-with-alt-tab

PR #336, opened 2017-07-18
Forces a window scale of 1 for Marco. Closes #335

### LINKED ISSUE #335: Marco should ignore GDK_SCALE so primitive HiDPI works
#### Expected behaviour

Exporting `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` into the Xsession should cause the entire MATE Desktop and it's suite of applications to correctly pixel double suitable for use on a computer equipped with a HiDPI display. 

#### Actual behaviour

When `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` are exported to the Xsession most of the MATE Desktop and it's suite of applications do scale correctly. However Marco demonstrates several issues where window and screen dimensions are not correctly calculated when these environment variables are exported. 

It is possible for coerce Marco to behave *"correctly"* when `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` are exported to the Xsession by creating the shell script `/usr/local/bin/marco` the overrides `GDK_SCALE` and `GDK_DPI_SCALE`, here is the script:

```bash
#!/usr/bin/env bash
env GDK_SCALE=1 GDK_DPI_SCALE=1 /usr/bin/marco $@ &
```

Therefore Marco should ignore/override `GDK_SCALE` to ensure operates correctly regardless of what the user has exported to the Xsession.

Export `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` into the Xsession, this can be done by editing `/etc/profile.d/mate.sh` on Ubuntu MATE 17.10 daily. Reboot and your desktop will be 2x scaled but Marco dimension and screen calculations are incorrect. 

#### MATE general version

MATE Desktop 1.18

#### Package version

Marco 1.18.1

#### Linux Distribution

Ubuntu MATE 17.10 daily.

@muktupavels understands this domain well and the hack above was created based on an IRC conversation we had some weeks ago. Tagging @monsta @XRevan86 and @ikeydoherty as they may be interested in this issue.

A similar bug was raised against `mate-panel`:

  * https://github.com/mate-desktop/mate-panel/issues/621

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47294644-marco-should-ignore-gdk_scale-so-primitive-hidpi-works?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flexiondotorg:
The pull request forces a specific window scale for `marco`, instead of using the default or user configured scale. 

Tested on a Dell XPS 15 with 3840 x 2160 and `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` exported to the Xsession. Windows and screen dimensions are now correct.

### DIFF
--- a/src/ui/ui.c
+++ b/src/ui/ui.c
@@ -62,6 +62,11 @@ void meta_ui_init(int* argc, char*** argv)
 	{
 		meta_fatal ("Unable to open X display %s\n", XDisplayName (NULL));
 	}
+
+  /* We need to be able to fully trust that the window and monitor sizes
+   * that GDK reports corresponds to the X ones, so we disable the automatic
+   * scale handling */
+  gdk_x11_display_set_window_scale (gdk_display_get_default (), 1);
 }
 
 Display* meta_ui_get_display(void)


lukefromdc:
How would I test this? I cannot risk exporting a scale factor to my entire desktop because I could have trouble getting back out of it and then find all the icons moved. I don't see how to test a WM in GtkInspector either. Also, the only way I know how to start an xsession with any exported env variable is from the profile files and that sort of thing.

flexiondotorg:
For those of you without HiDPI displays, just a confirmation there are no regressions will suffice ☺️

_Sent from my OnePlus ONEPLUS A5000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #339, opened 2017-07-29
avoid deprecated gdk_display_get_screen

sc0w:


raveit65:
I am just testing...

lukefromdc:
I just tested this both with and without compositing, could not find any behavioral differences. I normally use compiz though, so I could miss something subtle. The alt-tab and workspace switcher OSD worked fine, tested them.

raveit65:
Thank you.

PR #342, opened 2017-08-08
Fix synthetic keybinding/button-grab window

vkareh:
When a client is passively grabbing keys or mouse clicks that it does not need, it sends them up for other clients to process (this sets the `send_event` flag inside `XEvent`).

Often in this situation, the event contains the wrong window (either root, for global keybindings, or the original client itself). This means that Marco will attempt to process the event for the wrong window.

This is not an issue for global bindings within Marco, as the focused/current window does not matter. However, for shortcuts that operate directly on specific windows, the events gets lost.

This change addresses this issue by determining which is the currently-focused (for key grabs), or under-the-pointer (for button grabs), window, regardless of which client forwarded the event.

### DIFF
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -1779,6 +1779,22 @@ static gboolean event_callback(XEvent* event, gpointer data)
       meta_display_process_key_event (display, window, event);
       break;
     case ButtonPress:
+      /* Use window under pointer when processing synthetic events from another client */
+      if (window == NULL && event->xbutton.send_event)
+        {
+          int x, y, root_x, root_y;
+          Window root, child;
+          guint mask;
+          MetaScreen *random_screen;
+          random_screen = display->screens->data;
+          XQueryPointer (display->xdisplay,
+                         event->xany.window,
+                         &root, &child,
+                         &root_x, &root_y,
+                         &x, &y,
+                         &mask);
+          window = meta_display_lookup_x_window (display, child);
+        }
       if ((window &&
            grab_op_is_mouse (display->grab_op) &&
            display->grab_button != (int) event->xbutton.button &&
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -1785,8 +1785,6 @@ static gboolean event_callback(XEvent* event, gpointer data)
           int x, y, root_x, root_y;
           Window root, child;
           guint mask;
-          MetaScreen *random_screen;
-          random_screen = display->screens->data;
           XQueryPointer (display->xdisplay,
                          event->xany.window,
                          &root, &child,
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -1310,6 +1310,14 @@ meta_display_process_key_event (MetaDisplay *display,
       meta_ui_window_is_widget (screen->ui, event->xany.window))
     return;
 
+  /* Use focused window when processing synthetic events from another client */
+  if (window == NULL && event->xkey.send_event) {
+    Window focus = None;
+    int ret_to = RevertToPointerRoot;
+    XGetInputFocus (display->xdisplay, &focus, &ret_to);
+    window = meta_display_lookup_x_window (display, focus);
+  }
+
   /* window may be NULL */
 
 #ifdef HAVE_XKB


raveit65:
Can you provide a test case please?
Sorry at the moment a lot PRs to MATE projects are filed out.
A test case would speed up the process if i don't have to think about it how a PR can be confirmed to work.
Thx

lukefromdc:
I have no idea in the world how to test this

flexiondotorg:
I will review this, since I have a test case which is using `mate-hud` to activate when just Alt is depressed and released, which requires this patch.

vkareh:
Yes, of course! Another test case is to bind Maximize window to `<Super>Up` while having mate-menu (or brisk-menu) bound to `<Super>`.

Also, I'm pretty sure you'll need the most recent mate-settings-daemon, which I'm not sure is currently in the daily build (the latest tag doesn't include the fix, AFAIK)

raveit65:
> Yes, of course! Another test case is to bind Maximize window to <Super>Up while having mate-menu (or brisk-menu) bound to <Super>.

Ok, got it
I use the super+... key to start a lot of applications, for this reason i always disable the super key in mate-menu/mintmenu
Personal i think developer are to be shot if they add a single super key without a second key to their apps ;-)

But wasn't there already a commit/PR somewhere to fix that?
Or is this a dejvu..... :-)

vkareh:
@raveit65 - there is a fix for this on `mate-settings-daemon` that takes care of no swallowing KeyPress events in case you use the Keyboard Shortcuts app to set up a global keybinding. Marco needs it for window keybindings

vkareh:
@flexiondotorg - keep in mind that part of the test case for `mate-hud` also includes `<Alt>ButtonPress` with the window to make sure you can move it around (Button 1), or resize it (Button 2).

Also, breaking `<Alt>Tab` was a thing, which should also be taken care of with this PR :stuck_out_tongue:

muktupavels:
@flexiondotorg did you do code review? Or you just checked if it works?

Simply I think that metacity / marco should not start to process random events that are sent to root window for focused windows or windows that are under mouse.

For example Alt+F10 that is used to maximize / unmaximize window. What would you expect if some app send that to root window? Nothing, right? But after this patch focused window will be maximized...

vkareh:
Re-posting here from https://bugzilla.gnome.org/show_bug.cgi?id=786021 to include others in the discussion:

> I guess I don't see this as a bug as much as a feature request. In your example, I would expect an app simulating an Alt+F10 to actually maximize the focused window. It would allow global keybindings (which live in the root window anyway, from what I've seen) to allow acting on behalf of other non-root windows (like the HUD in my example)
> 
> Unless there's another way of doing this which I don't know about (which is entirely possible - I'm not that well versed in how to use X yet). Try as I might, I haven't been able to do this from the HUD. The event just gets ignored.

monsta:
So it seems the Metacity discussion ended in discovering some issue in mate-hud?
Sorry, I don't have enough knowledge of window managers, so I might not understand some things.

On the other side, looks like this change is needed to complement https://github.com/mate-desktop/mate-settings-daemon/pull/179 so that mintMenu/mate-menu could be bound to Super key without affecting the rest of the keybindings out there.

vkareh:
@monsta - the larger issue here is that we're trying to convince X that it's ok to use modifiers as global keybindings, but if a user hits `Alt` or `Super` the program (mate-hud or mate-menu respectively) will grab the keyboard. If the user subsequently hits another key (as for `Alt+F10`, for example), then mate-hud would need to forward that event to the window manager so that it can take care of it (maximizing in this example). However, since mate-hud already grabbed the Alt, then it read the F10, Marco is not aware of what window it was supposed to be sent to. This patch lets Marco identify which window it was intended to go to.

So @muktupavels is right. And this patch _could_ be avoided by replaying X events on the respective programs (mate-hud, etc) but then global keybindings end up sending strange characters to the focused window, which causes all sorts of bugs everywhere... So far I haven't found a more elegant solution than this... 😕

Until then, this patch seems to be necessary.

muktupavels:
> then mate-hud would need to forward that event to the window manager so that it can take care of it

mate-hud does not need to forward event to window manager; it needs to forward event to correct window.

> However, since mate-hud already grabbed the Alt, then it read the F10, Marco is not aware of what window it was supposed to be sent to.

Macro is not aware of that event, mate-hud stole it, macro does not see that event, same with application. If you (mate-hud) are not interested in that event you should deliver that event to window it was supposed to be delivered in first place.

@vkareh why global keybindings should send strange characters? What bugs?

vkareh:
@muktupavels - how do I know which window was interested in it in the first place? `event.child` always points to whichever window the mouse is on (regardless of sloppy focus settings), so that's not it. Knowing the answer to this question would solve a lot of headaches for me.

Yeah, with `ReplayKeyboard` in mate-hud, if I'm focused on the terminal and then hit `Alt+Tab`, it works, but also sends a Tab character to that same terminal (same with pretty much any window). If I do `Alt+F1` to bring up the global menu, it opens the menu, but also the Help window (F1) of whichever program I'm focused on, etc. Bindings like Alt+F10 and Alt+1/2/3, that operate on the currently focused window, work just fine with `ReplayKeyboard`, but global bindings like Alt+F1, Alt+Tab, send characters to the focused window.

Without the `ReplayKeyboard` (or in `GrabModeAsync`), it's the opposite: global bindings work perfectly well, but window bindings (Alt+F10, Alt+1/2/3, etc) need to be executed twice before they take... This patch to marco seems to solve that problem for any bindings that marco controls (Alt+F10, for example).

My hope is that I'm just doing the "modifier-only" detection wrong on mate-hud/mate-menu/brisk-menu and that there's some intelligent way of capturing it without affecting how "normal" keybindings work, but if there is, I haven't found a way. I combed through the Unity source code and it does it by having the window manager (Compiz) itself take care of all XKeyGrabs (including modifier-only ones for the HUD and the menu thing)... Plasma does the same directly from KWin (its window manager)... these options are a no-go for Marco/Metacity, since they shouldn't be concerned with bindings for external progams (Unity/Plasma are monolithic, MATE/Flashback are modular).

I guess at this point I'm looking for help in solving this issue with modifier-only keybindings - after weeks of trying different strategies I seem to be stuck with the current state. It seems that without support from either the window manager (or perhaps the settings daemon), modifier-only keybindings are not properly supported for apps. I hope I'm wrong, but I'm still going to play with m-s-d just in case.

Thoughts?

monsta:
Ok, what if we temporarily forget about mate-hud? What is needed for mintMenu/mate-menu?

vkareh:
@monsta - same deal. If you don't have this patch in marco, and you have a `Super` binding (for example, `<Super>Right` to tile the window to the right, or `<Super>c` to open the calculator, you'd need to press `<Super>c+c` for it do it. The first `c` is lost... If you then try to solve it by making the key grab synchronous and issuing a ReplayKeyboard, `<Super>Right` works well, but `<Super>c` sends a `c` character to the window (and then opens the calculator when pressing `c` a second time)

This patch solves this issue for mate-menu

monsta:
The calculator binding actually works for me, it must be from mate-settings-daemon where it's fixed by https://github.com/mate-desktop/mate-settings-daemon/commit/9eb0130a902aa509a3f4f61d0312af807fa87936.

ericmarceau-rogers:
I'm sorry, but I'm not up to speed with "correct vocabulary" as everyone else.

Can someone tell me if my reported problem [**Width of "grab-zone" for window edges/corners is too small (1 pixel)**](https://github.com/ubuntu/yaru/issues/4104) is

- related to this issue, and
- if so, why I am still encountering the issue **IF** it is being reported as fixed ?

Is there anything else that I need to do to make my issue "go away" or is the issue such that, as someone indicated on the UbuntuMATE site in a [**different posting**](https://ubuntu-mate.community/t/window-adjust-grab-zone-size-issues/28070/2?u=ericmarceau), that the issue is **inherent to Marco's design**?

If the latter, I don't understand why we would be forced to a adopt a compositor, just to have that wider edge detection zone?  

Since the issue is **Window-centred**, I can't visualize **why** a compositor should ever come into the picture for such edge detection!  

To me, the basic Window object should be able to offer/control its own properties (_and maybe I just don't know enough to know that that is not possible_).

But that is just me voicing my issue, from a User standpoint, without being a Linux developer.

Thank you in advance for your guidance/clarification.

lukefromdc:
The invisible grab area is the window shadow area, which is in fact
part of the window. The window drawn by the compositor is a slightly
larger window than the visible application window. It is enlarged by the
borders around it, known as GtkFrameExtents. The background of these
borders is always transparent, thus only a compositor can draw them
and a noncompositing window manager simply draws the content window
and decorations.

It is on these borders that window shadows are drawn,
and the side of them controls the available space for grabbing the 
window borders from just outside of them to resize. Moving a window
by dragging interacts with the titlebar not the shadow area so is the
same with and without compositing

PR #347, opened 2017-08-29
Allow tiled windows to be resized horizontally

### LINKED ISSUE #250: Windows are not resizable after snapping
Steps to reproduce:
1) Enable window snapping if it's not
2) Move some window to the left or right screen corner so it's snapped
3) Move the cursor to center of the screen and try to catch and drag the window corner, to expand it across the width.
Expected result: window is resized by mouse.
Real result: window stay with the same size.

The same result with Alt+F8. Snapped windows are not resizable.
Tested in Linux Mint 17.3 and Ubuntu MATE 16.04, both with MATE 1.12
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30331563-windows-are-not-resizable-after-snapping?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
Determine whether the user is initiating a resize action on a tiled
window. If the user is trying to grab the window for resizing
horizontally from the edge that's farther away from the screen edge,
allow the resize to occur. Otherwise maintain the current tile geometry.

Also modified the window hints to allow resizing from the window menu.

Fixes #250

### DIFF
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -876,6 +876,7 @@ constrain_tiling (MetaWindow         *window,
   gboolean hminbad, vminbad;
   gboolean horiz_equal, vert_equal;
   gboolean constraint_already_satisfied;
+  gboolean allow_resize = FALSE;
 
   if (priority > PRIORITY_TILING)
     return TRUE;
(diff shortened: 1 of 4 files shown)


raveit65:
That works and it doesn't break 'toggle back to previous size' if you using a keybinding for tiling support, which happens with other PR to fix open report.
But if i resize the window in tiling mode and click on the metacity window decoration, the window snap back to default tiling size.
Is that a new feature *restore-back-to-default-tiling-size* ?  :-)

vkareh:
@raveit65 - good catch, I didn't even notice that :stuck_out_tongue_closed_eyes:

I'll remove that "feature" ;)

vkareh:
@raveit65 - fixed it! Also noticed that the same bug prevented Firefox from resizing properly - this last commit also fixes that

raveit65:
lol..... firefox behaviour is like a rubber band :-)

raveit65:
Thanks.

meequz:
Thank you very much guys!

monsta:
@vkareh: can you please have a look at https://github.com/mate-desktop/marco/issues/356 and https://github.com/mate-desktop/marco/issues/357?

vkareh:
@monsta - of course! I'll take a look

Issue #353, opened 2017-10-13
Marco memory leak


CliveMcCarthy:
#### Actual behaviour Marco grabs memory and does not free it. It can accumulate GB's of memory
I switched to Compiz and the problem goes away.

lukefromdc:
https://github.com/mate-desktop/marco/commit/6cf1369769b94165c3360945deb47621e957c603
https://github.com/mate-desktop/marco/pull/358
was merged well after the release of 1.18, which came out in March 2017.  Was it ever backported to 1.18?

vkareh:
1.18 is no longer supported and the fix has been commited since 1.19.2. Closing...

PR #358, opened 2017-11-08
Fix #329: Compositer leaks XFIXES regions

### LINKED ISSUE #329: X11 resource leak causes high Xorg CPU usage and sluggish UI
After a login session is more than a few days old (currently 8 days), X consumes more and more CPU time and the user interface becomes very sluggish.

Investigating Xorg with gdb showed a lot of time was being spent in GetXIDRange which is known to be inefficient for large numbers of resources.

xrestop showed marco as the larges resource consumer:
0 - marco ( PID: 2772 ):
        res_base      : 0x3000000
        res_mask      : 0x1fffff
        windows       : 24
        GCs           : 51
        fonts         : 2
        pixmaps       : 34
        pictures      : 52
        glyphsets     : 2
        colormaps     : 0
        passive grabs : 1780
        cursors       : 0
        unknowns      : 288940
        pixmap bytes  : 130435987
        other bytes   : ~6982424
        total bytes   : ~137418411

The "unknowns" resource seems to be constantly growing.  Even when the screen is locked.

This behavior was observed on Ubuntu Mate edition version 16.04.2 LTS, marco package version 1.12.1-1

I can't imagine anything unique to my environment which would cause this so I'm very surprise this hasn't been reported in the past.

Restarting marco restores proper performance

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46527985-x11-resource-leak-causes-high-xorg-cpu-usage-and-sluggish-ui?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

J-Bu:
XFIXES regions are leaked during paint_windows if more than one META_COMP_WINDOW_DESKTOP is available (#329)

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -1216,8 +1216,15 @@ paint_windows (MetaScreen   *screen,
 
           if (cw->type == META_COMP_WINDOW_DESKTOP)
             {
-              desktop_region = XFixesCreateRegion (xdisplay, 0, 0);
-              XFixesCopyRegion (xdisplay, desktop_region, paint_region);
+              if(desktop_region)
+              {
+                XFixesUnionRegion (xdisplay, desktop_region, desktop_region, paint_region);
+              }
+              else
+              {
+                desktop_region = XFixesCreateRegion (xdisplay, 0, 0);
+                XFixesCopyRegion (xdisplay, desktop_region, paint_region);
+              }
             }
 
           XFixesSubtractRegion (xdisplay, paint_region,


raveit65:
Well, i don't have a server runing 24-7 and i shutdown my box in the evening?
Do i see an improvement directly after restarting marco after adding the i commit to my build?
Maybe a general reduce of memory consumption of marco?

raveit65:
I can confirm that memory consumption is much lower after a restart of marco, so it's true that we have a leak.
Now marco is running with your fix, let see if memory will increase again....

raveit65:
After a day of testing i saw not really a different.
Steps to reproduce please?

J-Bu:
Hi,

to trigger the leak you need at least two windows with window type [desktop](https://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472629520). Then you can see with [xrestop](https://www.freedesktop.org/wiki/Software/xrestop/) that marcos use of Misc resources will increase on each window draw event (if anything changes on the screen). If the Misc resources hits a critical value something arounf 150000-170000 you will see that XOrg is using 100% cpu on one core looping through his resource array.

I don't think you will see a create increase in memory usage.

J-Bu:
I wrote a caja desktop replacement which uses one window per monitor but that is not published yet.

raveit65:
merged https://github.com/mate-desktop/marco/commit/6cf1369769b94165c3360945deb47621e957c603
Thanks

monsta:
Can this also fix https://github.com/mate-desktop/marco/issues/353, or is that issue different?

Issue #218, opened 2015-09-12
Add more tiling options


ciechomke:
I think that tilling windows to left and right border is not enough. Look at tilling in KWin - by default you can tile windows to right and left borders and also to corners so it expands number of tilled windows at once to four. It also provides more configuration options to tile more than 4 windows like in tilling WM's (awesome for example). Many people use MATE Desktop to work (code, managing remote hosts) due to it's lightweight and responsibility so this feature can improve suitability of MATE to work with many windows at one workspace.
##

milankni:
That is especially usefull on large screens, the window on half screen tiling is still unnecessarily too large.

jean:
My screen is exactly right for three terminals. Two terminals each taking up half the screen are too wide.

GiedriusS:
Similar issue: https://github.com/mate-desktop/marco/issues/194. Perhaps this one or the other one could be closed?

raveit65:
What compiz-0.88 can since a long time is simply genius.
9 positions on screen selectable via keyboard shortcuts +
3 sizes for one position selectable if you press the same shortcut again.

reydun:
This would really be appreciated, I'm seriously contemplating switching away from Marco to openbox just for this. Compiz can do this, but it's too resource heavy for my taste.

I really hope someone is working on this for a future release at least.

raveit65:
> This would really be appreciated, I'm seriously contemplating switching away from Marco to openbox just for this. Compiz can do this, but it's too resource heavy for my taste.

> I really hope someone is working on this for a future release at least.

We are a small theme and we do our best to make users happy, but a day has only 24 hours ;)
Patches are welcome.
Hope you will use marco again if it is done.

v1kn:
What is the current state of implementing this feature?  

To many users corner tiling is a crucial functionality and a deal braker, and since MATE DE has been getting traction, popularity, and a lot of work is being done (looking at ubuntu mate), I thought I would bring back some attention to this. 

Looking forward to this feature, please consider it

monsta:
Corner tiling is implemented now (was also requested at https://github.com/mate-desktop/marco/issues/101). You will get it with 1.20 release.
Not sure if anyone would implement tiling for more than 4 windows on the screen (like 9 positions in compiz mentioned above).
Should we close this or keep it open?

flexiondotorg:
Keybindings are available in MATE to tile to more than just quadrants and quadrant tiling is available via window dragging in 1.20. This can be closed.

PR #359, opened 2017-12-02
Implemented corner tiling

ozeidan:
Hi,
this is my attempt at implementing the corner tiling feature requested in issues #218. 
It works fairly well already when dragging the window with the mouse. I tried implementing the keyboard shortcuts for this but I didn't manage to understand how I can get the shortcuts to show up in gsettings. A little help on this would be appreciated.

### DIFF
--- a/src/50-marco-window-key.xml.in
+++ b/src/50-marco-window-key.xml.in
@@ -78,6 +78,22 @@
 	_description="Tile window to west (left) side of screen" />
 
 	<KeyListEntry
+	name="tile-to-corner-nw"
+	_description="Tile window to north-west (top left) corner" />
+
+	<KeyListEntry
+	name="tile-to-corner-ne"
+	_description="Tile window to north-east (top right) corner" />
+
+	<KeyListEntry
+	name="tile-to-corner-sw"
+	_description="Tile window to south-west (bottom left) corner" />
+
+	<KeyListEntry
+	name="tile-to-corner-se"
+	_description="Tile window to south-east (bottom right) corner" />
+
+	<KeyListEntry
 	name="move-to-corner-nw"
 	_description="Move window to north-west (top left) corner" />
 
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -206,6 +206,11 @@ static inline void get_size_limits       (const MetaWindow        *window,
                                           MetaRectangle *min_size,
                                           MetaRectangle *max_size);
 
+static void can_resize_tiled         (guint tile_mode,
+                                          int resize_gravity,
+                                          gboolean *resize_horizontally,
+                                          gboolean *resize_vertically);
+
 typedef gboolean (* ConstraintFunc) (MetaWindow         *window,
                                      ConstraintInfo     *info,
                                      ConstraintPriority  priority,
@@ -876,13 +881,14 @@ constrain_tiling (MetaWindow         *window,
   gboolean hminbad, vminbad;
   gboolean horiz_equal, vert_equal;
   gboolean constraint_already_satisfied;
-  gboolean allow_resize = FALSE;
+  gboolean allow_resize_horizontally = FALSE;
+  gboolean allow_resize_vertically = FALSE;
 
   if (priority > PRIORITY_TILING)
     return TRUE;
 
   /* Determine whether constraint applies; exit if it doesn't */
-  if (!META_WINDOW_SIDE_TILED (window))
+  if (!META_WINDOW_TILED(window))
     return TRUE;
 
   /* Calculate target_size - as the tile previews need this as well, we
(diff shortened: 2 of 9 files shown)


ozeidan:
So I managed to get the keybindings running. Here is a small preview:
corner-tiling
I will continue to look through the code once more because I am not sure if I am causing any side effects with this change.

ozeidan:
@raveit65 I applied a quick fix for this, please try it out again. 

Currently, when a window is tiled into a corner it is maximized_vertically, which intuitively doesn't make much sense, but technically it seems to be the intended behaviour. Is that fine to leave in?

ozeidan:
@lukefromdc I also noticed this behavior with Skype Preview, I'll try to fix this.

lukefromdc:
I don't know what anyone else uses, but I have compiz setup to corner tile windows dragged to corners and ignore the top, bottom, left, and right edges entirely as far as tiling is concerned. It would seem to me that the only reason to use the corner as a target for dragging a window to tile it would be to tile the window quarter-screen from that corner.

ozeidan:
@lukefromdc what are you trying to say? Isn't how you describe it the exact current behaviour? 

I am not able to 100% reproduce the bug with some windows not tiling to corners. Some hints to this would be appreciated.

lukefromdc:
I normally do not use tiling with Marco at all because I move windows around a lot (and have jittery hands) and tiling any window on the edges acts in a way I find hard to control. The suggestion that windows should tile to halfscreen(maximized vertically) rather than quarterscreen on being dragged to a corner is what I was objecting to and could not understand being the intended behavior.

raveit65:
I am pretty sure some users will point with the finger on it, if vertical resizing doesn't work.

The issue with dconf-editor i can't reproduce.
Here the preview and the tiling is working well, tested with f26 and dconf-editor-3.23.4-2.fc26.x86_64.
Same behaviour with other gnome apps like gnome-abrt, gnome-mpv, gitg.
The only issue i see with csd windows is that resizing a tiled window don't work.
But csd windows are using invisible borders for resizing and not a decorator edge,
This affects also our side-by-side-tiling, so this is a complete  other story and not related to PR.
I recall we have already a report about and it's save to ignore it for the moment, imo.

>  I normally do not use tiling with Marco at all because I move windows around a lot (and have jittery hands) and tiling any window on the edges acts in a way I find hard to control.

For me only the top edge is problematic because of the resizing to max window size, which can be a unwanted result if you move a window, especially with small notebook monitors.
But that's another story too :-)

lukefromdc:
I am using an old version of deconf-editor (3.18) because I prefer its UI

ozeidan:
Hi, so again I changed quite a few things. The resizing should work properly now. There might still be bugs, so please test as much as you can.

vkareh:
Awesome - latest commit fixes the unmaximization issue for me - it works great!

I was wrong about the CSD windows - they tile just fine, except for `dconf-editor`, as @lukefromdc mentioned... not sure why that is, but everything else is behaving correctly.

Code looks good to me as well :)

ozeidan:
I checked the issue with dconf-editor. It seems like the culprit is that the height of the size_hints of the dconf window is smaller than the height we want to resize it to when corner tiling. If I understand it correctly, those come from the application?

raveit65:
@ozeidan 
Are you familiar with the 'git rebase command?
It looks like some commits can squashed together.
```
git rebase -i <last-good-commit>
```
-i means interactive and it opens an editor (vi), you will see how it works.
I think you know it better than me how much commits you need.

Btw. i don#t have the issue here with  dconf-editor-3.23.4-2.fc26.x86_64
So, i am thinking it's a problem with d-e-3.25

raveit65:
@lukefromdc 
Do you like to test more?

ozeidan:
@raveit65 So if my assumption is true, tiling dconf-editor (or other programs) may work on some screens but not on others. On bigger screens, the height of a 'corner' might be bigger than the minimum height that dconf-editor supplies. 

I renamed and squashed some commits on my local branch. Can I just force push them?

raveit65:
@ozeidan 
yeah, i forgot to say. git push -f will update yor repo and the PR automatic

raveit65:
@ozeidan 
You're right with dconf-editor. I works well on my my main box with a 27“ monitor, but not with the same rpm on my notebook with a (17,3“) Widescreen Display.

raveit65:
merged
https://github.com/mate-desktop/marco/commit/b471b912d68a05711a95a3c0e033c0577262f18e
https://github.com/mate-desktop/marco/commit/8828bd7cabc39f0a3564467c2303774dc494fd0b
https://github.com/mate-desktop/marco/commit/01f68e8f9f1541393e6e069db08f834d4ef18691
https://github.com/mate-desktop/marco/commit/6219f8e8bcaeefb9185a3c3f5f20de4e2fa8f18f
https://github.com/mate-desktop/marco/commit/d8ef314fda4caf147162817bae4274910e9abc31
Thanks a lot for this feature.

Issue #314, opened 2017-04-16
[Feature Request]: Show workspaces when moving windows with the keyboard? [$15]


orschiro:
Dear all, 

This request originated from the [Ubuntu MATE](https://ubuntu-mate.community/t/show-workspaces-when-moving-windows/12668/2) community.

I love the keyboard shortcuts to switch workspaces and to move the active window across different workspaces. 

However, while I do see on which workspace I am when switching workspaces, I lack this indication when moving the active window across the workspaces. 

Such as indication would be incredibly helpful to place the active window at the appropriate workspace. 

What are your thoughts?

Yours,

Robert

orschiro:
Show this when moving windows across workspaces:

![show workspace switch](https://cloud.githubusercontent.com/assets/170947/25452278/5bd39c38-2ac5-11e7-96af-6a9e71a56ff5.png)

orschiro:
Any further thoughts or opinions on this request?

samtygier:
Sounded simple enough so I had a look at the code. Looks like more than an afternoon project, so here are my notes in case anyone else has time to pick this up before I get back to it.

The popup indicator is called TabPopup, and lives in tabpopup.c.

Workspace switching and window moving are somewhat similar. The respective keybindings call handle_switch_to_workspace() or handle_move_to_workspace(). The can be relative (left, right) or absolute (go to desktop x). switch shows the TabPopup on relative switches. Move only switches desktop on relative moves (flip=True). So it looks like the smart thing to do is to merge the 2 functionalities into say handle_switch_and_move_to_workspace(), which can take an additional bool to say if the window is moving.

This works for the first movement. But if you keep the SHIFT+CTRL+ALT held down, then you stay in the grab, and everything is done by meta_display_process_key_event() calling process_workspace_switch_grab(). This seems to reimplement various things and does things with modifier masks that I have not figured out yet.

ozeidan:
Hi,
I implemented this, see PR #366.
Could you please try if the patch works on your machine, @orschiro? Thanks

orschiro:
@ozeidan thanks so much for your PR! 

I don't have MATE installed at the moment and cannot test. 

Anyone else?

TomaszGasior:
Patch works great! Like it, without issues.

orschiro:
Now it just needs to be merged and you can claim your bounty, @ozeidan :-)

sc0w:
[...]

PR #366, opened 2017-12-20
Implemented tabpopup when moving window to different workspace

ozeidan:
Hi again guys,
I implemented moving a window to a neighbouring workspace 'properly', it's now similar to how the workspace switching is implemented. This allowed me to make the tabpopup also show up when moving as you can see in this gif:

tabpopup

This solves issue #314. 

I believe it works fairly well, the only unfinished thing being how I handled getting the correct MetaKeyBindingAction in file keybindings.c at line 2742, which is a small hack. For the window move keybindings (up, down, left, and right) this call returned values from the range 94-97 which is why I hardcoded the new enum variants in prefs.h at line 199 like this:
```

...
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_LEFT = 94,
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_RIGHT = 95,
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_UP = 96,
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_DOWN = 97
 } MetaKeyBindingAction;

```
But I am pretty sure this only works on my machine. A little help on this would be appreciated.


raveit65:
This PR (with the hack) works well here on my system and i like that the moved window looks different than other windows in tabpopup.
With 36 workspaces and several horizontal lines the tabpopup looks a bit like oldschool tetris now, when moving a window :-)
Sorry, i can't help with the hack, i am more a specialist in gtk3 themes and rpm package maintainer in fedora.
But i will call the other guys.
@monsta @lukefromdc @XRevan86 @vkareh @sc0w 
Can you please help?

TomaszGasior:
Works great!!! I really like it and I hope it will be part of next MATE release. Thanks, ozeidan.

sc0w:
@ozeidan 

In `keybindings.c`, on the function:
```

handle_workspace_switch_or_move  (MetaDisplay    *display,
                                  MetaScreen     *screen,
                                  MetaWindow     *window,
                                  XEvent         *event,
                                  MetaKeyBinding *binding,
                                  guint           is_move)
```

why **guint** ?

I see you handle that value with `TRUE` and `FALSE`, why not **gboolean** ? (and the indentation is wrong there)

@lukefromdc 

you can move the active window to another workspaces with:
`[control] + [shift] + [alt] + [arrows]`

lukefromdc:
Turns out that works in compiz-reloaded too, giving a rather flashy effect if combined with the rotating cube. Thus this PR should be good for behavioral consistancy between Marco and Compiz, the two WM's most likely to be used in MATE.

ozeidan:
@sc0w you are right, I changed that.

lukefromdc:
Merged with 
https://github.com/mate-desktop/marco/commit/9e2b7eff713b6881a60538e08e222dad0d364647

raveit65:
@ozeidan @lukefromdc @sc0w @monsta 
...
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_LEFT = 94,
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_RIGHT = 95,
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_UP = 96,
+  META_KEYBINDING_ACTION_WORKSPACE_MOVE_DOWN = 97
 } MetaKeyBindingAction;

```
Guys, can this hardcoding be a problem with non-western or a exotic keyboard layout?

lukefromdc:
I have no easy way to test that, as I could easily disable getting back to a layout I can actually use. We can revert this if an issue is likely, or simply remove the hardcoded keybindings with another commit unless all the default keybindings are hardcoded. Are these keybindings users can reset like most of them?

lukefromdc:
I just realized the flags from libmatekbd can be used to recover using only the mouse and was able to play with this. I looked a a whole bunch of keyboard arrangements, could not find one where these particular keys got any change beyond some name variations. I then added Persian Kurdish and Afghani layouts and found that moving a window with these in Marco in either of these worked exactly the same. This does not mean no keyboard arrangement where these are different exists, but I was not able to find one that broke this in a short test.

PR #369, opened 2017-12-29
Fix for unfullscreening after resizing tiled window

### LINKED ISSUE #357: Marco 1.19.1: Tiled window doesn't restore its size after fullscreen, if it has been resized
Marco 1.19.1: Tiled window doesn't restore its size after fullscreen, if it has been resized. It restores the size properly, if the window hasn't been resized. I also made a video to show the problem: https://www.youtube.com/watch?v=qW1ClbadH7I

#### Expected behaviour
Marco remembering the window's size/dimensions even if it has been resized after tiling, so that it can restore them after this window exits fullscreen.

#### Actual behaviour
The window's dimensions are not remembered and the window size becomes equal to the whole screen.

Use a window that can go easily fullscreen (like browser on youtube, or vlc). Then tile the window to one edge, resize it just a bit. Make it fullscreen and then exit fullscreen.

#### MATE general version
1.19.1

#### Package version
marco 1.19.1

#### Linux Distribution
Manjaro

ozeidan:
With this fix the window gets correctly set back to its tiled state after leaving fullscreen mode and after having resized it when it was tiled. Please test

Fixes #357

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -2902,6 +2902,7 @@ meta_window_make_fullscreen_internal (MetaWindow  *window)
       meta_window_save_rect (window);
 
       window->fullscreen = TRUE;
+      window->tile_resized = FALSE;
       window->force_save_user_rect = FALSE;
 
       meta_stack_freeze (window->screen->stack);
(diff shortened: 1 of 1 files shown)


raveit65:
some spaces are removed and merged
https://github.com/mate-desktop/marco/commit/7e2aea33a6fa5ba3bed740b65aaafb872551c150
Thanks

raveit65:
@ozeidan 
Do you think it is possible to add an gsettings key to disable/enable tiling to maximize window state from top edge of the monitor?
I ask because i know from some some ppl that it is very annoying if you move a window with the mouse and touch the top edge by accident, than the window goes to a unwanted max. state.
Especially with small notebook monitors.
For this reason i disabled tiling feature in fedora with a gsetting override.
But with corner tiling it looks like it's time to enable it, but without the annoying top screen tiling as default.

Issue #334, opened 2017-07-15
Context menu of GTK headerbar


TomaszGasior:
#### Actual behaviour
Marco is not integrated with GTK3 headerbars.

#### Expected behaviour
Marco should provide context menu for windows with client side decorations.

#### Steps to reproduce the behaviour
1. Start xfwm4, XFCE4 window manager instead of Marco.
2. Run app without CSD and right-click on the titlebar — you will see window manager menu.
3. Run any app with CSD (dconf-editor for example) and right-click on empty area of headerbar — you will see the same menu as in 2. point.
4. Start Marco.
5. Run app with CSD (dconf-editor for example) and right-click on headerbar — you will see fallback context menu from GTK instead of Marco context menu.

#### Package version
any Marco version

monsta:
As https://github.com/mate-desktop/marco/pull/370 is merged, this can be closed.

PR #370, opened 2017-12-29
Use the native window menu for windows with Gtk3 CSDs

vfscanf:
Gtk3 applications that utilize the GtkHeaderBar widget (basically windows with CSD) only show a very minimal "fallback" window menu when right clicking on the title bar. Gtk3 introduced an API that allows the window manager to display the native window menu, even on CSD windows.

To do that we have to register a X11 Atom. Then, when somebody right clicks on the CSD title bar, we get a client event, which we can handle appropriately.

xfwm4 already implements this, here is the relevant patch: https://mail.xfce.org/pipermail/xfce4-commits/2015-February/059211.html . There is also a bug report for this feature #334 .

This PR adds support for that API. I have tested it on my system, and it works fine so far.


TomaszGasior:
Works great for me with marco from master.  Thanks, vfscanf!

vfscanf:
There is one small issue that I found myself. When the title bar from a window is offscreen, the window menu shows an additional option "Move Titlebar On screen". The check that decides whether to show that option doesn't work for CSD windows, as they have no titlebar from the viewpoint of the window manager.

This check will always decide that the window is offscreen for CSD windows. As a result, this option is always shown for CSD windows, but it doesn't do anything. I will push another commit that fixes this faulty check so it doesn't show this option for CSD windows.

TomaszGasior:
@vfscanf, are you able to also fix this bug If user presses Alt+Space to open window context menu:
marco: 
alt-space-marco
xfwm4 (good position of context menu):
alt-space-xfwm4

raveit65:
@TomaszGasior 
Can you please open a report for it instead of hijacking this PR?
Thank you.

TomaszGasior:
it seems to me it's related to this PR.

raveit65:
Second commit removes the wrong entry.

raveit65:
> It seems to me it's related to this PR. When we are adding native menu to CSD-windows we can also fix its position at the same time.

Nope, the issue with wrong position of the menu when pressing alt+space on csd-windows still exists without using this PR.
So, please file out a new report for it, it is confirmed.

raveit65:
Thanks

Issue #28, opened 2012-12-15
firefox and thunderbird windows can't be moved with keyboard shortcuts


cheater:
Hi,
I'm not sure if I reported this here already, or if we just spoke about this in #mate. I wasn't able to find this report in the tracker anywhere.

I set up keyboard shortcuts in mateconf-editor under the marco app. They are for "move window to east edge" and "move window to west edge".

Those keyboard combos work for many applications, but don't for firefox. Initially the window assumes the new position, but as soon as I let go of the keyboard combination the window jumps back to where it was before.

I believe there were more programs that did that but I can't remember what they were off the top of my head.

Since Firefox is a fairly popular program and moving windows around is daily routine in two-monitor environments I would ask that this gets looked into. Currently Firefox makes up more than 50% of my work day and it would be a huge improvement in productivity for me if this started working properly. As it is I can only use FF on one monitor.

I am on Ubuntu 12.04.

$ marco --version
marco 1.4.1

$ firefox --version
Mozilla Firefox 18.0

Thanks
##

NiceandGently:
Is the issue still exits with Mate-1.8/10 ?

monsta:
> Initially the window assumes the new position, but as soon as I let go of the keyboard combination the window jumps back to where it was before.

Goddammit, it's exactly like that in Marco 1.11. Only Firefox is behaving like that. Usual apps like Caja or mate-keybinding-properties have their windows moved like they should.

anorman728:
I'm having this issue on Ubuntu 16.04, with MATE installed through apt-get instead of using Ubuntu MATE.  MATE version is 1.12.1.

raveit65:
Confirmed with marco-1.19.0-1.fc26.x86_64.
But it works with metacity and gnome keybindings.

cheater:
@monsta looks like you only tried it with programs provided by mate... try it with programs not provided by mate.

geoffm:
I'm having this issue on a fresh install of Ubuntu MATE 16.04

vkareh:
This is not an issue in MATE 1.22, as far as I know. Is anyone else experiencing this or can we close?

PR #371, opened 2017-12-30
Multimonitor keybinds

### LINKED ISSUE #320: Multi-monitor keyboard shortcuts [$25]
Currently (as of 1.12.1, not sure about newer versions), there aren't any keybaord shortcuts that can be set to move the current window to the next/previous monitor. I'm constantly forced to use my mouse in order to move the window.

This is something I'm really missing from other window managers, and multi-monitor setups have become very commonplace nowadays (compared to the gnome2 days), so I'd really like to see this.

<bountysource-plugin>

---
There is a **[$25 open bounty](https://www.bountysource.com/issues/45282538-multi-monitor-keyboard-shortcuts?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ozeidan:
Hi,

I implemented proper keybinds for moving windows between monitors that work with tiled/maximized/fullscreen windows.

Please test it and sorry for the PR spam :)

Closes #320

### DIFF
--- a/src/50-marco-window-key.xml.in
+++ b/src/50-marco-window-key.xml.in
@@ -130,6 +130,22 @@
 	_description="Move window to center of screen" />
 
 	<KeyListEntry
+	name="move-to-monitor-n"
+	_description="Move window to north (top) monitor" />
+
+	<KeyListEntry
+	name="move-to-monitor-s"
+	_description="Move window to south (bottom) monitor" />
+
+	<KeyListEntry
+	name="move-to-monitor-e"
+	_description="Move window to east (right) monitor" />
+
+	<KeyListEntry
+	name="move-to-monitor-w"
+	_description="Move window to west (left) monitor" />
+
+	<KeyListEntry
 	name="move-to-workspace-1"
 	_description="Move window to workspace 1"
 	value="1"
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -800,7 +800,7 @@ constrain_maximization (MetaWindow         *window,
 
   /* Determine whether constraint applies; exit if it doesn't */
   if ((!window->maximized_horizontally && !window->maximized_vertically) ||
-      META_WINDOW_SIDE_TILED (window))
+      META_WINDOW_TILED (window))
     return TRUE;
 
   /* Calculate target_size = maximized size of (window + frame) */
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -2670,6 +2670,47 @@ handle_move_to_center  (MetaDisplay    *display,
           window->rect.height);
 }
 
+
+static void
+handle_move_to_monitor  (MetaDisplay    *display,
+                         MetaScreen     *screen,
+                         MetaWindow     *window,
+                         XEvent         *event,
+                         MetaKeyBinding *binding)
+{  
+  MetaScreenDirection move_direction = binding->handler->data;
+  MetaXineramaScreenInfo* current;
+  MetaXineramaScreenInfo* neighbour;
+  MetaRectangle current_window_rect;
+  int new_x;
+  int new_y;
+  
+  
+   current = meta_screen_get_xinerama_for_window(screen, window);
+   neighbour = meta_screen_get_xinerama_neighbor(screen, current->number, move_direction);
+
+   if(neighbour == NULL ||
+      current->number == neighbour->number)
+     return;
+
+   if(META_WINDOW_TILED (window))
+     {
+       window->tile_monitor_number = neighbour->number;
+       return;
+     }
+
+   meta_window_get_client_root_coords(window, &current_window_rect);
+
+   new_x = current_window_rect.x - current->rect.x + neighbour->rect.x;
+   new_y = current_window_rect.y - current->rect.y + neighbour->rect.y;
+   /* target_rect.width = window->rect.width; */
+   /* target_rect.height = window->rect.height; */
+   /* Maybe do some resizing? */
+   
+   meta_window_move(window, TRUE, new_x, new_y);
+   
+}
+
 static gboolean
 process_workspace_switch_grab (MetaDisplay *display,
                                MetaScreen  *screen,
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -2678,37 +2678,20 @@ handle_move_to_monitor  (MetaDisplay    *display,
                          XEvent         *event,
                          MetaKeyBinding *binding)
 {  
-  MetaScreenDirection move_direction = binding->handler->data;
-  MetaXineramaScreenInfo* current;
-  MetaXineramaScreenInfo* neighbour;
-  MetaRectangle current_window_rect;
-  int new_x;
-  int new_y;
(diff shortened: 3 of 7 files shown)


jhasse:
How can  I assign those keybindings when running a local marco build via `./src/marco --replace`?

ozeidan:
@jhasse @lukefromdc you have to recompile the gschema: I usually do this by copying the file `src/org.mate.marco.gschema.xml` into `/usr/share/glib-2.0/schemas` and then `running lib-compile-schemas /usr/share/glib-2.0/schemas`. Then the options `move-to-monitor-(n|e|s|w)` should show up in dconf-editor at path `/org/mate/marco/window-keybindings/`.

ozeidan:
Also, I still found an issue with this implementation: When moving a firefox window that is either fullscreen or maximized to another monitor, it jumps right back. Can you guys reproduce that?

lukefromdc:
I did recompile the schema, and the default disabled keybindings showed up in dconf-editor, but when I tried things like <ALT>F12 or <ALT>m they had no effect. I'm not an expert on this, my workflow is mostly mouse driven with few keybindings used and only the ctrl-alt-backspace changed back to its old binding to kill the X server/

ozeidan:
@lukefromdc west means right to left. Try placing the window on the right monitor and then pressing the key.

As a side note: currently I am using a function that already existed to determine the neighbour in the according direction. This works fairly well, though the current algorithm stops checking once it has found a monitor in the given direction. This means that if you have two monitors on a side of the current monitor, one the one which was registered first will always be chosen. Maybe I should improve this logic by checking which of those the screens is closer vertically (in the cases of move-to-monitor-(w|e)). What do you guys think?

raveit65:
> Also, I still found an issue with this implementation: When moving a firefox window that is either fullscreen or maximized to another monitor, it jumps right back. Can you guys reproduce that?

Yes, it's reproducible here, but only with firefox.  I tested this with chrome, vlc, mate-terminal and i don't see this issue.
firefox problem?

>  Maybe I should improve this logic by checking which of those the screens is closer vertically (in the cases of move-to-monitor-(w|e)). What do you guys think?

I don't think that i can test a 3 monitor setup. There is a report somewhere that it is impossible to setup 3 Monitors with mate-display-properties.
And i don't have 2 external monitors/tvs in one room ;-)
But go ahead if you can test it. I am sure there are some ppl who use 3 or more monitors.

Setting up the keybindings is working with both tools here, mate-keybinding-properties and dconf-editor.
Only keys with more than 2 levels are critical. An example for move-to-monitor-north with a vertical monitor setup.
```
crtl+Shift+Alt+KP_Up    --> works
crtl+Shift+Alt+u             --> works
crtl+Shift+Alt+8             --> don't work
```
The last one is impposible to setup with german keyboard layout and mate-keybinding-properties and move-to-monitor-north function doesn't work.
The result is always
```
crtl+Shift+Alt+[
```
With german layout you need *altgr* key to type *[* 

(I hope this is understandable because of my poor english)

I think such weirdness is more a general problem than a problem with your PR.

raveit65:
@lukefromdc 
Could you solve to setup some working keys?
@ozeidan 
Beside from firefox, do you think this is ready to go?
We want to freeze MATE in a few days for releasing 1.20 end of this month, than only bugfixes are allowed to merge.

ozeidan:
The only other issue I found was, when you have a fullscreen window on a monitor and move a window to that monitor, the moved window gets placed behind the fullscreen window. I'll try to fix this today or tomorrow, otherwise I think it's ready to merge.

raveit65:
Ok, but isn't a fullscreen-window not always in the foreground?
I never saw another window ontop of a fullscreen-window.
Or do i miss something?

monsta:
Looks like https://github.com/mate-desktop/marco/pull/371/commits/15da872d741a683ba57d38ed08f6c42608447f14 might affect some existing features.

monsta:
> When moving a firefox window that is either fullscreen or maximized to another monitor, it jumps right back.

I'm not surprised at all. We already have https://github.com/mate-desktop/marco/issues/28. Maybe Mozilla needs to learn to use GTK+?

lukefromdc:
I have zero experience setting up custom keybindings, and also would need to be damned sure they got reverted. Not sure if I will have time for this one today.

lukefromdc:
I have no issue with merging this

ozeidan:
@raveit65 when a window is fullscreen, you still can alt-tab to get another window in front of it. So I think if you move a window to that monitor it should be on top, just because it's really confusing otherwise. 

@monsta I changed the parameter from FALSE to TRUE because I found some bug that this fixes. As I remember it this only changes the behavior of the move-to-X keybinds

raveit65:
> when a window is fullscreen, you still can alt-tab to get another window in front of it. So I think if you move a window to that monitor it should be on top, just because it's really confusing otherwise.

Indeed, i didn't known that, btw. i never tried this ;-)
The result is that the panel, in my case top and bottom panel are on top of the fullsreen-window,
which looks very scary if you use a transparent panel.
I don't think i will use that for productivity :-)

But, i think that both arguments are legitim , for me the alt-tab function is a special case with a fullscreen window.
Any way, i don't care much about if this is an issue or not.

monsta:
What is that bug? Is it in this PR or in some existing feature?

raveit65:
@ozeidan 
Today is the last day for merging before feature freeze for 1.20.
Otherwise  this new feature goes into next development cycle (1 year)
Personal i can live with the fullscreen-window issue for the moment.
If i merge the PR today you can fix it later.

ozeidan:
@monsta I believe the bug was, when moving a window with move-to-side-X so
that it switches to another monitor, the user_rect would stay on the old
monitor (because that parameter was false). Then fullscreening it, it would
pop up back on the old monitor. I noticed this because the keybindings
introduced in this PR also had this issue initially. The other change is
something that I forgot to change in a previous PR, although it doesn't
change any behavior. It's just an inconsistency in the code which came to
be because if the corner tiling system. In hindsight I shouldn't have made
these undocumented changes, sorry for that.

@raveit65 I think the PR is good to go for now. I'll try to look at the
other bug this week, but I don't think it's very critical

raveit65 <notifications@github.com> schrieb am Mo., 15. Jan. 2018, 16:38:

raveit65:
Thanks.

csarmendariz:
Hi, 

Sorry because I'm pretty new so the question can be a bit silly...
I would love to be able to move windows between monitors, I think is the only functionality I'm missing in marco.
However I cannot find how to activate the keybinds you guys are talking about here... not sure if I have an old version of marco without this functionality activated.
I'm running ubuntu mate 16.04, is there a way I could activate a keyboard shortcut to send window to the other monitor?
Thanks and sorry if this is not the place to ask.

raveit65:
You need Mate-1.20 (ubuntu.18.04)

PR #373, opened 2017-12-31
Implemented a setting to disable top tiling

ozeidan:
@raveit65 

I implemented a setting to disable the maximization when dragged to the top of the screen.
You have to recompile the gschema file to test it :)

The two settings are now called `allow-tiling` and `allow-top-tiling`

### DIFF
--- a/src/core/prefs.c
+++ b/src/core/prefs.c
@@ -120,7 +120,8 @@ static gboolean resize_with_right_button = FALSE;
 static gboolean show_tab_border = FALSE;
 static gboolean center_new_windows = FALSE;
 static gboolean force_fullscreen = TRUE;
-static gboolean side_by_side_tiling = FALSE;
+static gboolean allow_tiling = FALSE;
+static gboolean allow_top_tiling = TRUE;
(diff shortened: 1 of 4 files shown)


ozeidan:
I noticed that I forgot to handle the 'shaken-loose' option. That means, when a window is already maximized, and you pull it down and back up to the window border, it will still maximize back, despite the setting being disabled. Should the setting also disable this behavior?

raveit65:
> Should the setting also disable this behavior?

No, this happens only if the window is already maximize and during draging.
Here the moving direction is away from top edge, so i don't think a unwaned window maximizing can happen....,normal....
I never heard that a user complains about.
Thanks a lot for adding this key, i will test it soon.

flexiondotorg:
@raveit65 I agree with you.

lukefromdc:
Fine here too

raveit65:
merged and a bit reworded https://github.com/mate-desktop/marco/commit/709fe3eb2f8987ffc598ecbece4997acdf51cdca
Thanks again

PR #374, opened 2018-01-01
Feature/window move tabpopup correction

ozeidan:
I figured out how the order of the enum is determined: it depends on the order of the keybinds in the all-keybindings.h. So I moved the move-to-workspace keybinds after the the switch-to-workspace keybinds in both the all-keybinds.h file and the enumeration, which is arguably more elegant. 

@raveit65 could you do another small test? If it works I think it can be merged as nothing else has changed


raveit65:
merged https://github.com/mate-desktop/marco/commit/52701a81d82cd89525ed883cb046f11f384c8d6c
Thanks again

PR #384, opened 2018-02-12
Fixes move_to_center moving window between monitors

ozeidan:
When two or more monitors are connected, the move_to_center keybind moves the window to the center of the whole xinerama, which in case of two monitors means that the window lands between the monitors. I think that this is unwanted behavior (?). This is a small fix that moves the window to the center of the monitor that the window belongs to.


vkareh:
Thanks for fixing this - works great!

PR #387, opened 2018-02-15
avoid deprecated GtkStock

sc0w:


muktupavels:
The commit message is misleading, you are not avoiding deprecated GtkStock... You just changed from function that produce warnings to g_object_new that does not do that, but button still use GtkStock.

https://developer.gnome.org/gtk3/stable/GtkButton.html#GtkButton--use-stock
Don't remember how, but there are way to enable warnings for deprecated properties. Here is commit that add that info to docs:
https://gitlab.gnome.org/GNOME/gtk/commit/5c3c95a96cf09f3f25605e1d7b8796b644c75ac4

I would say you have made situation worse than it was if you will ever want to port to gtk 4. Removed function would fail to compile, this will result in run time warning about property that does not exist that will be harder to debug.

If you want to shut up warning then just use `G_GNUC_BEGIN_IGNORE_DEPRECATIONS` and `G_GNUC_END_IGNORE_DEPRECATIONS`. Or if you want to really avoid GtkStock then you need to use icon names.

Commit that removes GtkStock:
https://gitlab.gnome.org/GNOME/gtk/commit/590cd6a9f2df789839429c1bd77f84d23c074047

muktupavels:
You probably will get button with `gtk-ok` text on it.

lukefromdc:
OK, I just confirmed thats exactly what happens: if "use-stock" is changed to "FALSE" you get no icon in the button and the text reads "gtk-ok"

sc0w:
@muktupavels thanks for the info

I avoided the warning: depecated GtkStock

I thought this is the correct way because I can't see build warnings

I will review all the PRs related merged

Gtk4 is our enemy :)

sc0w:
Fixed in https://github.com/mate-desktop/marco/pull/389

monsta:
@muktupavels: maybe you know - are pre-defined button text names (like "Open", "Find", etc.) gone along with stock items as well? I'm asking because the changes like https://github.com/mate-desktop/marco/pull/389, https://github.com/mate-desktop/eom/pull/160, https://github.com/mate-desktop/mate-control-center/pull/333 add new strings for translation to our code, while the same strings were already translated in GTK+.

muktupavels:
@monsta I think that is expected. See:
https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-22/gtk/deprecated/gtkstock.c#L329

These strings are translated inside GTK+, under `gtk30` domain. It internally use `g_dgettext` to get translated text. You are using `gettext("_Open")` / `_("_Open")` to get translated text, that automatically use domain that is set with `textdomain`. It does not magically start looking under other domains if it does not find text in current domain.

So if you want to access translated string from gtk+ you would use something like this `g_dgettext ("gtk30", "_Open")`, but not sure if that is good idea. What if GTK removes one or more translated strings in future?

monsta:
Yeah, I guess it's not the way we want to go. Though it was nice to have all items with the same text translated in one place...

Issue #383, opened 2018-02-11
No visual difference for unfocused CSD windows


bwat47:
Expected behaviour
When a window is inactive there should be a visual difference compared to an active window (e.g. different font color in the titlebar)

Actual behaviour
This works as expected for non-csd apps, but for csd apps (e.g. gnome disks) there is zero visual difference between a focused and un-focused window

NOTE: I'm not sure if marco even has any control over this or not... (as these are csd apps). However, I noticed that if I switch from Marco to Compiz, it does work as expected with compiz (the un-focused gnome disks window gets a faded font color in the titlebar)

Steps to reproduce the behaviour
1. Install gnome disks (or any other gnome app with a csd)
2. open another window next to gnome disks and put focus on that new window

MATE general version
1.20

Package version
Marco 1.20

Linux Distribution
Solus

bwat47:
For comparison:

Compiz: https://i.imgur.com/rMvHrED.png

Marco: https://i.imgur.com/81XA2UB.png

monsta:
What you see is a problem in a window manager theme. Try another theme.

bwat47:
just to more clearly illustrate what I mean, non-csd windows always work as expected (see screenshot), the problem is specific to windows using csd decorations.

this is the case which each theme I've tried (including any of the default 
![windows](https://user-images.githubusercontent.com/1442520/36400396-a380ee36-159e-11e8-92e5-077a6588bf07.png)
mate themes)

For any given theme I've tried, it *does* work as expected with csd apps when the theme is used under compiz or under gnome-shell.

lukefromdc:
I can confirm this here with marco 1.20, AND with compiz-reloaded build against marco 1.20 and GTK 3.22.27. In GtkInspector I am finding that neither the :background nor the :focus state is coming up on CSD windows, thus theme changes for these are not applied

monsta:
Ok, I guess it's in Marco then. Wonder if Metacity is affected too.

lukefromdc:
Metacity, IceWM, and compiz 0.9 branch are NOT affected, 
Marco and compiz 0.8 branch ARE affected, the :backdrop state flag is not set so themes do not respond

raveit65:
old known issue, the **.backdrop** state is still broken with marco.
On the other side @flexiondotorg told me to disable this state in ubuntu themes a while a go.
To avoid this weird theming with compiz in ubuntu-mate, not everyone like this......
But i agree this should work with marco in general.
[...]

PR #392, opened 2018-02-22
window: add _NET_WM_STATE_FOCUSED hint to _NET_WM_STATE

lukefromdc:
Fixes https://github.com/mate-desktop/marco/issues/383 if :backdrop state supported in themes
ported from:
https://github.com/GNOME/metacity/commit/4ccb99a50c54f345c4c7d9ac77f1ea76bc6c7c74

### DIFF
--- a/src/core/atomnames.h
+++ b/src/core/atomnames.h
@@ -158,6 +158,7 @@ item(_NET_WM_ACTION_ABOVE)
 item(_NET_WM_ACTION_BELOW)
 item(_NET_WM_STATE_STICKY)
 item(_NET_WM_FULLSCREEN_MONITORS)
+item(_NET_WM_STATE_FOCUSED)
 
 #if 0
 /* We apparently never use: */
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -1259,7 +1259,7 @@ static void
 set_net_wm_state (MetaWindow *window)
 {
   int i;
-  unsigned long data[12];
+  unsigned long data[13];
 
   i = 0;
   if (window->shaded)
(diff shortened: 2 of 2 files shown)


raveit65:
I just tested this with Adwaita and BlueMenta with that commit reverted.
http://git.mate-desktop.org/mate-themes/commit/?id=32a139134af1231418f6b6db631eb16b703dd787
Don't expect too much with BlueMenta, especially with our own apps.
For CSD apps you see some differences.
All this theme code is old and needs a rework, btw. we need a clear concept for should be differ in backdrop state.

lukefromdc:
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy/commit/b490e97c166fbca95a3d6fdc404fea0ea1bcae61

includes code on how to use this

vkareh:
Also, no regressions here. I use the `gtk3-nocsd` library. One thing I noticed that might be theme related is that whatever widget is focused on the window also changes. Here are two screenshots, notice the selected line in dconf how the text changes from white (focused) to black (not focused). This is Ubuntu MATE with the Ambiant MATE theme
focused
not-focused

lukefromdc:
Many GTK themes deliberately change the look of non-toplevel widgets when the toplevel window is in backdropp. This may be the intended behavior, remember that Ubuntu used compiz 0.9 (with this feature) for years and now is one GNOME

vkareh:
Makes sense, thanks!

raveit65:
> This is Ubuntu MATE with the Ambiant MATE theme

@vkareh 
Most .backdrop code lines in ubuntu themes are disabled by a wish of @flexiondotorg 
to avoid the styling with compiz.
Still wondering that you found a setting :)

raveit65:
@lukefromdc @sc0w 
Ok, let's start to adjust themes.
This is a bugfix, but I would prefer to fix themes first before we push that to a stable branch.
On the other side this is big visual change if this landed in an installation. So, maybe it's better to leave that for 1.22, and distro maintainer can decide for their own to use it as patch before next stable release.

PR #399, opened 2018-03-21
Support HiDPI window decorations

### LINKED ISSUE #386: [1.20.0] [HiDPI] marco doesn't scale with window-scaling-factor
#### Expected behaviour

Marco should scale according to `org.mate.desktop.interface.window-scaling-factor` so that close, maximize, minimize, etc. match the interface DPI.

#### Actual behaviour

Marco appears as if it is using a scale of 1 (original size), which could be different than the desired scale factor.

1. Acquire an installation of MATE on a HiDPI monitor.
2. Set the `window-scaling-factor` to a value larger than `1` (i.e. `2` to double the interface scale).
3. Click apply; observe no change in the scale of the window frame on certain applications (e.g. MATE Terminal, Caja, among others).

#### MATE general version

`1.20.0`

#### Package version

`master` branch using Gentoo eBuild `>=mate-base/mate-9999`

#### Linux Distribution

Gentoo Linux

### LINKED ISSUE #393: [HiDPI] Windows not properly maximised on second screen of different resolution
#### Expected behaviour

Windows should maximise normally... To use all available screen estate, regardless of which screen the windows happen to be on.

#### Actual behaviour

See the video below... (laptop resolution 2560x1440, external screen resolution 1920x1080)

[video.tar.gz](https://github.com/mate-desktop/marco/files/1754072/video.tar.gz)

Set `window-scaling-factor=2`, and move a window to the other screen, and attempt to maximise the window (see video)

#### MATE general version

MATE 1.20.0.

#### Package version

mate-* 1.20.0.

#### Linux Distribution

AOSC OS, packaging tree: https://github.com/AOSC-Dev/aosc-os-abbs/tree/staging/extra-mate/.

N/A

vkareh:
* Fixes #386 
* Fixes #393

### DIFF
--- a/src/ui/frames.c
+++ b/src/ui/frames.c
@@ -524,10 +524,12 @@ meta_frames_calc_geometry (MetaFrames        *frames,
                            MetaUIFrame       *frame,
                            MetaFrameGeometry *fgeom)
 {
-  int width, height;
+  int width, height, scale;
   MetaFrameFlags flags;
   MetaFrameType type;
   MetaButtonLayout button_layout;
+
+  scale = gdk_window_get_scale_factor (frame->window);
 
   meta_core_get (GDK_DISPLAY_XDISPLAY (gdk_display_get_default ()), frame->xwindow,
                  META_CORE_GET_CLIENT_WIDTH, &width,
@@ -544,9 +546,14 @@ meta_frames_calc_geometry (MetaFrames        *frames,
                             type,
                             frame->text_height,
                             flags,
-                            width, height,
+                            width / scale, height / scale,
                             &button_layout,
                             fgeom);
+
+  fgeom->top_height *= scale;
+  fgeom->bottom_height *= scale;
+  fgeom->left_width *= scale;
+  fgeom->right_width *= scale;
 }
 
 MetaFrames*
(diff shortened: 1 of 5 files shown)


flexiondotorg:
I've rebuilt Marco with these patches. Window controls are scaled appropriately. However I've noticed that regardless of scale factor tiling a window disregards struts.

lukefromdc:
I just wish this had also make compiz-reloaded's decorations that depend on Marco scale for hidpi. No such luck, that will require some coding there.

raveit65:
@flexiondotorg 
Do you see issues and like to see changes or why you didn't approved it ?

flexiondotorg:
I did observe one regression, but after further testing suggests this patch is not the cause.

  * When I toggle `window-scaling-factor` the desktop is redrawn and any panels are missing. To recover panels I have to use MATE Tweak to switch panel layouts to restore panels.
  * This issue is reproducible using Marco and Compiz 0.9. I'll try and track down the culprit.

raveit65:
>   * When I toggle `window-scaling-factor` the desktop is redrawn and any panels are missing. To recover panels I have to use MATE Tweak to switch panel layouts to restore panels.

I think the redraw of the panel is done with calling "killall mate-panel" from a previous HIDPI commit, (maybe m-s-d).
So this works only if the panel was started by the session-manager.

tidux:
Well that's dumb.  SVG should -always- look good at any resolution.

PR #400, opened 2018-03-24
Scale cursor size for HiDPI

### LINKED ISSUE #388: marco: pointer does not obey Gtk scaling at start
#### Expected behaviour
On a HiDPI system, the cursor should be scaled up as well.

#### Actual behaviour
The cursor stays annoyingly small until the size is adjusted.

Go into Appearance -> Customize -> Pointer and slide the cursor size slider all the way to the right and back.  This forces the pointer to resize and obey scaling.  This happens regardless of which theme is applied.

#### MATE general version
1.20.

#### Package version
1.20.

#### Linux Distribution
Arch Linux

n/a - upstream problem

vkareh:
Fixes #388

### DIFF
--- a/src/core/prefs.c
+++ b/src/core/prefs.c
@@ -27,6 +27,7 @@
 #include "prefs.h"
 #include "ui.h"
 #include "util.h"
+#include <gdk/gdk.h>
 #include <gio/gio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


flexiondotorg:
Works for me on the XPS 15 with HiDPI display.

PR #401, opened 2018-03-29
GSettings preference to control compositor shadow rendering

chrisnovakovic:
Please don't merge yet: help/advice needed (see below).

This is an (almost complete) patch that allows for the compositor's rendering of drop shadows to be controlled via a new GSettings preference, `org.mate.Marco.general.compositing-shadows`: a value of `true` enables shadows, a value of `false` disables them.

What this doesn't yet do:

* Existing windows aren't automatically rerendered with/without shadows as appropriate when the preference's value is changed. I know I need to add a new branch to the conditional in `prefs_changed_callback()` in `src/core/display.c` that checks whether changes were made to `META_PREF_COMPOSITING_SHADOWS`, but I'm not familiar enough with Marco to know what exactly needs to be done in that case. (I've already tried a few things that didn't work.)
* Shadows are still rendered beneath tooltips when the preference's value is set to `false`. (This also seems to be the case even when the `META_DEBUG_NO_SHADOW` environment variable is set, so this might actually be a separate issue unrelated to this patch.)
* An appropriate checkbox needs to be added to *Window Preferences → General → Compositing Manager* to control the preference in a more user-friendly way. I can write that patch against mate-control-center when this patch is complete.

Even with this patch, shadows are still rendered beneath GTK+3 windows with client-side decorations, but of course there's nothing Marco can do about that.

Any help with fixing items 1 and 2 in the list above would be appreciated!

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -38,6 +38,7 @@
 
 #include <cairo/cairo-xlib.h>
 
+#include "prefs.h"
 #include "display.h"
 #include "../core/display-private.h"
 #include "screen.h"
@@ -2786,7 +2787,9 @@ xrender_manage_screen (MetaCompositor *compositor,
   info->overlays = 0;
   info->clip_changed = TRUE;
 
-  info->have_shadows = (g_getenv("META_DEBUG_NO_SHADOW") == NULL);
+  info->have_shadows = (g_getenv("META_DEBUG_NO_SHADOW") == NULL &&
+                        meta_prefs_get_compositing_shadows() == TRUE);
+
   if (info->have_shadows)
     {
       meta_verbose ("Enabling shadows\n");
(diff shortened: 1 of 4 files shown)


raveit65:
Why should someone use the compositor if shadows are disabled?
I mean this is the main visible attribute of a compositor.

lukefromdc:
Other benefits of a compositor are ability to display rounded corners on a toplevel window, and ability to use transparency. Shadows require transparency but are not the only use of it. Much can be done with it in theming.

vkareh:
Good point, there's also a few smaller things like rendering window thumbnails when Alt+Tabbing

raveit65:
@chrisnovakovic 
Any progress or can we close this?

chrisnovakovic:
I haven't made any progress since opening the PR and without external help it's unlikely I'll finish it, I'm afraid.

flexiondotorg:
@vkareh Can you offer some assistance on the one? Looks like a potentially useful feature.

lukefromdc:
Any more on this or should we close it?

chrisnovakovic:
I don't use MATE any more - feel free to close it if you don't think this is worth implementing.

PR #404, opened 2018-04-24
Correct resize popup position and reported size on HiDPI displays

vkareh:
When resizing certain types of windows (e.g. mate-terminal), there is a overlay popup that reports the current size of the window. On HiDPI displays, this popup is offset with respect to the window, and the size reported is relative to device pixels. This change fixes both issues so that the popup is in the correct location with the correct number of application pixels.


src/core/xprops.c
```diff
@@ -894,8 +899,9 @@ size_hints_from_results (GetPropertyResults *results,
   hints->min_height = cvtINT32toInt (raw->minHeight);
   hints->max_width  = cvtINT32toInt (raw->maxWidth);
   hints->max_height = cvtINT32toInt (raw->maxHeight);
-  hints->width_inc  = cvtINT32toInt (raw->widthInc);
-  hints->height_inc = cvtINT32toInt (raw->heightInc);
+  /* Convert the increments to application pixels */
+  hints->width_inc  = cvtINT32toInt (raw->widthInc) / scale;
+  hints->height_inc = cvtINT32toInt (raw->heightInc) / scale;
```
muktupavels: Why only `width_inc` and `height_inc`? If these are scaled then I would expect that other things are scaled too... It just seems weird to keep mixed things in one struct, either you unscale all or nothing.

vkareh:
@muktupavels - I changed this now. Instead of scaling the window hints, now it just determines whether to show the resize pop-up if the increment hints are larger than the scale factor. This means that all geometry calculations that rely on the hints will be as reported by X, so no weird side effects.

monsta:
How to get that popup showing for something else besides mate-terminal? I've never seen it anywhere else.

raveit65:
> How to get that popup showing for something else besides mate-terminal? I've never seen it anywhere else.

This is exactly the issue if you use 
```
[rave@mother ~]$ gsettings get org.mate.interface window-scaling-factor
2
```
You should see it with a non-hidipi monitor too, if you resize caja.
But the popup is outside of caja window.
Maybe you will see it in a VM, where you can scroll the desktop if the desktop is larger than VM window.

vkareh:
@monsta - yes, the offset is set by Marco, you would need to be running Marco with GDK_SCALE=2 as well (or just set `window-scaling-factor` to `2` and restart Marco)

monsta:
Ah, right. Ok, so this PR has no effect if I just restart Marco without these scale variables. If I restart it with them, then popup indeed disappears for any scaled apps that I run after that.

raveit65:
A far i know those values in mate-terminal popup are lines numbers, same sizes what you see in menu terminal.
So, this doesn't make any sense for other applications outside of VTE.

monsta:
> How to get that popup showing for something else besides mate-terminal? I've never seen it anywhere else.

So, if it's not possible to see the popup (except for terminal windows) without HiDPI, then this should be good to go. The behavior is now the same with and without HiDPI, so it's consistent.

PR #405, opened 2018-04-25
Fix frame geometry to avoid stuck resize cursor

vkareh:
This seems to fix the resize cursor being stuck in the wrong image. I went through my checklist from when I added the scaling to marco and couldn't find any regressions, but please test to make sure it works for you.

### DIFF
--- a/src/ui/frames.c
+++ b/src/ui/frames.c
@@ -549,11 +549,6 @@ meta_frames_calc_geometry (MetaFrames        *frames,
                             width / scale, height / scale,
                             &button_layout,
                             fgeom);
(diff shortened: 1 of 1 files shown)


raveit65:
@lukefromdc 
This only reproducible with a HIDIPI monitor, see my little video.
https://www.dropbox.com/s/77mld2f15ct9l4h/xcursor.ogv.tar.xz?dl=0

monsta:
This should have no impact on non-HiDPI systems (as `scale` is 1), so if it fixes some issues for you, merge it.

PR #407, opened 2018-05-09
ui: replace deprecated gtk_menu_popup

raveit65:
Please test.

### DIFF
--- a/src/ui/menu.c
+++ b/src/ui/menu.c
@@ -100,28 +100,6 @@ static MenuItem menuitems[] = {
 	{META_MENU_OP_DELETE, MENU_ITEM_IMAGE, MARCO_STOCK_DELETE, FALSE, N_("_Close")}
 };
 
-static void popup_position_func(GtkMenu* menu, gint* x, gint* y, gboolean* push_in, gpointer user_data)
-{
-	GtkRequisition req;
-	GdkPoint* pos;
-
-	pos = user_data;
-
-	gtk_widget_get_preferred_size (GTK_WIDGET (menu), &req, NULL);
-
-	*x = pos->x;
-	*y = pos->y;
-
-	if (meta_ui_get_direction() == META_UI_DIRECTION_RTL)
-	{
-		*x = MAX (0, *x - req.width);
-	}
-
-	/* Ensure onscreen */
-	*x = CLAMP (*x, 0, MAX(0, WidthOfScreen (gdk_x11_screen_get_xscreen (gdk_screen_get_default ())) - req.width));
-	*y = CLAMP (*y, 0, MAX(0, HeightOfScreen (gdk_x11_screen_get_xscreen (gdk_screen_get_default ())) - req.height));
-}
-
 static void menu_closed(GtkMenu* widget, gpointer data)
 {
 	MetaWindowMenu *menu;
@@ -498,7 +476,7 @@ void meta_window_menu_popup(MetaWindowMenu* menu, int root_x, int root_y, int bu
 	pt->x = root_x / scale;
 	pt->y = root_y / scale;
 
-	gtk_menu_popup(GTK_MENU (menu->menu), NULL, NULL, popup_position_func, pt, button, timestamp);
+	gtk_menu_popup_at_pointer (GTK_MENU (menu->menu), NULL);
 
     if (!gtk_widget_get_visible (menu->menu))
       meta_warning("GtkMenu failed to grab the pointer\n");


src/ui/menu.c
```diff
@@ -498,7 +476,7 @@ void meta_window_menu_popup(MetaWindowMenu* menu, int root_x, int root_y, int bu
 	pt->x = root_x / scale;
 	pt->y = root_y / scale;
 
-	gtk_menu_popup(GTK_MENU (menu->menu), NULL, NULL, popup_position_func, pt, button, timestamp);
+	gtk_menu_popup_at_pointer (GTK_MENU (menu->menu), NULL);
```
muktupavels: `pt` is now unused...

raveit65: Should i really do that?
```
diff --git a/src/ui/menu.c b/src/ui/menu.c
index 01f09c5..9bbf51d 100644
--- a/src/ui/menu.c
+++ b/src/ui/menu.c
@@ -467,14 +467,7 @@ meta_window_menu_new   (MetaFrames         *frames,
 
 void meta_window_menu_popup(MetaWindowMenu* menu, int root_x, int root_y, int button, guint32 timestamp)
 {
-	GdkPoint* pt = g_new(GdkPoint, 1);
-	gint scale;
-
-	g_object_set_data_full(G_OBJECT(menu->menu), "destroy-point", pt, g_free);
-
-	scale = gtk_widget_get_scale_factor (menu->menu);
-	pt->x = root_x / scale;
-	pt->y = root_y / scale;
+	g_object_set_data_full(G_OBJECT(menu->menu), "destroy-point", NULL, g_free);
 
 	gtk_menu_popup_at_pointer (GTK_MENU (menu->menu), NULL);
```
It seems to be working in runtime...

muktupavels: `g_object_set_data_full` is also unneeded... It was used to free allocated `GdkPoint`.

raveit65: Hmm, i noticed that i already killed using `alt-spacebar` to open the window-menu with first commit.
Maybe because of using `current event` (NULL) with gtk_menu_popup_at_pointer ?

muktupavels: Right, if there is no `GdkEvent` then how GTK/GDK can know where to popup menu?

https://developer.gnome.org/gtk3/stable/gtk3-General.html#gtk-get-current-event
You can pass `NULL` if you know that `gtk_get_current_event` will return event.

I don't remember details, but I think keybindings does not use gtk/gdk... In `metacity` I changed that popup function parameters so I can pass `GdkEvent`. There are places where you have it, in other places I just created new event from x event data.

raveit65:
I tried to back port your commit from metacity but now the menu doesn't open any more under pointer (somewhere on the desktop)
https://www.dropbox.com/s/q9lmlm6qknukdrm/marco_0001-menu-avoid-deprecated-gtk_menu_popup.patch?dl=0
I think this part is probaly wrong as i had to do changes for myself.
window.c
```
@@ -5473,23 +5473,19 @@ meta_window_client_message (MetaWindow *window,
   else if (event->xclient.message_type ==
            display->atom__GTK_SHOW_WINDOW_MENU)
     {
-      gulong x_root, y_root;
-      guint32 timestamp;
-      int button;
+      GdkRectangle rect;
 
       if (meta_prefs_get_raise_on_click ())
         meta_window_raise (window);
 
-      timestamp = meta_display_get_current_time_roundtrip (display);
-      x_root = event->xclient.data.l[1];
-      y_root = event->xclient.data.l[2];
-      button = 3;
+      rect.x = event->xclient.data.l[1];
+      rect.y = event->xclient.data.l[2];
+      rect.width = 0;
+      rect.height = 0;
 
       meta_window_show_menu (window,
-                             x_root,
-                             y_root,
-                             button,
-                             timestamp);
+                             &rect,
+                             (GdkEvent *) event);
     }
 
   return FALSE;
```
I think values for rect are wrong :)

muktupavels:
Don't know what is wrong, you will have to debug it yourself! You have HiDPI monitor, it might be related to that... All I know is that in `metacity` it works.

raveit65:
Ok.

raveit65:
I forgot to say that the menu with metacity is unscaled with HIDPI here.

muktupavels:
I know, there are problems with HiDPI...

sc0w:
in gnome, it was done with this commit:

https://gitlab.gnome.org/GNOME/metacity/commit/37fa0d19f35a7fbff3638a6e6deb108500d41abf#11e4e98e99a9eaf7f9f060a6b06801d163c9dc24

PR #422, opened 2018-08-27
backport: metacity-theme-3

### LINKED ISSUE #105: support metacity-theme-3.xml
I think in windows with Global Dark Theme enabled must have dark header (if theme supports). Mutter do this.

Screen from Mutter:
2014-05-24-161117_1366x768_scrot

And from Marco:
2014-05-24-161152_1366x768_scrot
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2321737-feature-request-gtk3-dark-theme-support?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
This is a series of backported commits to implement metacity theme format 3. Currently supports up to version 3.4, which means Menta and BlueMenta now load v3 without errors.

This also adds a new gsettings option to attach modal windows (defaults to false), which I don't like, but is needed to support the correct theme versioning, otherwise loading a v3 theme makes marco spit out errors. If we want, we could keep the theme data so things don't break, but disable/hide that key if we don't want MATE to have attached dialogs - I find them very weird to use...

There's a lot more work for full metacity-theme-3, but this is the minimum necessary so that themes can load without crashing.

Fixes #105


lukefromdc:
What should be used as a test theme for this?

vkareh:
Both Menta and BlueMenta have a metacity-theme-3 version

vkareh:
@lukefromdc - yeah, not everything is supported any longer, but the idea is to enable this theme format so that we can eventually have invisible resize borders (right now the resize border is only as wide as the window border, which is 1px in many themes)... it's a looong backlog of backports to get there though, thanks for testing this

raveit65:
> Also, I noticed window borders were much narrower in the version 3 version of BlueMenta, not sure if that's deliberate or not but its quite visible if you know what to look for.

This was intended. I increase the border width to improve resizing windows in MATE with metacity-xml-2 version.
The metacity-xml-3 was only used by fedora-cinnamon a while ago before linuxmint themes supported gtk+-3.20.
Cinnamon support the so called 'invisible borders' from mutter/metacity to resize windows.
So, increasing the border width wasn't needed for metacity-xml-3.
If we switch to using metacity with invisible border support  i will revert the thick a bit ugly looking borders in our themes.
Honestly i am waiting for that......

raveit65:
> ....but the idea is to enable this theme format so that we can eventually have invisible resize borders (right now the resize border is only as wide as the window border, which is 1px in many themes)... it's a looong backlog of backports to get there though

Nice if you want to do that, but isn't it easier to use metacity as VM instead of back porting features from them?
We only need to adjust a few config tools to support metacity again.
And keep in mint that metacity is well mantained and has a perfect support, better as we ever can maintain marco ;)

raveit65:
Thanks for that, i am happy to use the new settings for tiled-windows.
Btw. what is a modal-(dialog) window?

vkareh:
@raveit65 - modal dialog is a [GtkDialog](https://developer.gnome.org/gtk3/stable/GtkDialog.html), which belongs to another window. If you enable the attach settings, those dialogs show up as attached to the parent window and cannot be moved away from them. This is a very gnome-y thing and I already found the backport that removes the behavior (while keeping the theme artifacts). Will open a new PR soon.

I agree that metacity is well maintained, but there are a few things it that break full compatibility with MATE, as well as slightly different feature set. If we update Marco with the latest from Metacity we can get good support for stuff, but we can also set the direction we want to go in the future (like we have for certain things, like tiling support, keybindings, etc).

rilian-la-te:
Do dark themes is supported now?

rilian-la-te:
I quite like MATE and use it every day, (except mate-panel), so, it is important for me)

raveit65:
> but there are a few things it that break full compatibility with MATE

Which features?
And metacity isn't used by gnome-shell, so i am in doubt that Albert will break compatibility with MATE. Also we're in daily contact with him.
And marco can be use as secondary backup WM.
So what should happen?
For me having 2 WMs which do basically the same is only waste of resources and developer time.
How much commits should we backport further from metacity to make people happy?
And claiming it as our work :)
Sounds like a never ending story.

vkareh:
@raveit65 - by your logic, why are we not using gnome-flashback instead of mate? I prefer marco, like I said, because features like keybindings and tiling support, which are not supported in metacity in the same way.

Also, there's a list of other features that I would like to implement that are not part of metacity. I could submit pull requests for metacity, but then it becomes a battle of purposes, where gnome-flashback dictates the behavior of our desktop, regardless of which direction mate wants.

By design, gnome-flashback will always be a session of gnome 3. Also, parts of it are used as fallback when things go wrong in gnome-shell, which means its development is tightly bound to gnome-shell. But mate, as much as we backport from gnome, is free to choose a completely different direction.

So yeah, we have a long list of things to backport, but I'm not backporting everything either, just the things I think marco could use and to support features I'd like to implement. It _is_ a never ending story, but that's no different from using GTK, which changes daily.

The only other suggestion I have is to re-fork metacity and re-apply our mate-specific changes, but that's probably just as much work...

vkareh:
@muktupavels - I thought gnome-flashback was loaded when mutter failed?

rilian-la-te:
Was, yes, but not now.

muktupavels:
@vkareh `gnome-flashback` has never been used as fallback! That is module that I created to provide functionality that was moved into mutter and/or gnome-shell... There was time when `gnome-panel` was used as fallback, but I think (might be wrong) that starting with GNOME 3.8 it no longer true.

GNOME Flashback == GNOME + gnome-flashback + gnome-panel + gnome-applets + metacity | compiz.

vkareh:
got it, that makes sense, thanks for clarifying

raveit65:
> by your logic, why are we not using gnome-flashback instead of mate?

Because using gnome-flash back means to use g-s-d, which means to use nautilus because background handling is different with g-s-d

> I prefer marco, like I said, because features like keybindings and tiling support, which are not supported in metacity in the same way.

Yes, mate-keyboard-properties needs to adjusted to support keybindings from metacity and tiling features from marco needs to be included to metacity.

But adding 'invisible borders', dark-themes, vulkan support. etc from metacity/mutter to marco looks like more work for me.

I maintaining MATE since 2012 and i had to learn the supporting a desktop is a 24/7  job. All founders and most devs have lost MATE  after 1-2 years, because of a lack of time doing this several years.
At least we're  a small team and we should bring Mate and flashback together to save resources and manpower with a synergy effect.
But if every single dev thinks we need our own apps than this will never happen.

Well, that's my opinion but maybe i am only tired of all that work for MATE after more than 6 years.

raveit65:
Btw . in early MATE days there was an invitation from Vincent Untz (former gnome-panel developer) to develop Mate under the hood of gnome.
Sadly, former Mate founders never answered to this invitation.
Gnome-flashback did jump in to this gap ;)

muktupavels:
`metacity`, `gnome-panel` and `gnome-applets` definitely could be used by both - GNOME Flashback and MATE...

vkareh:
So how do we collaborate on this then? We need to settle a few things then:

*Window manager*
I agree that metacity is more stable, so we would need to backport new marco features into metacity and make metacity the default window manager. How do we ensure that metacity development follows the mate ideal without breaking gnome flashback? Could we implement features in metacity that @muktupavels is not interested in, but that would benefit mate?

*Panels*
We would need to settle the in-process vs out-of-process debacle. Also, gnome-panel pulls in a lot of gnome dependencies, including nautilus. So we would need to retain mate-panel, except for that means deviating code-bases. So how do we keep collaborating in panels without breaking either mate or gnome flashback? Seems like here we need to keep our own for now...

*Applets*
We could conceivably merge these into a single repo, or split them all into individual repos, I'm fine with either...

If those are really the only components to gnome flashback, then this could be maneagable (apart from the dependencies aspect).

There's also the logistics: how do we merge the teams? does everyone want this? where do we host stuff? etc. etc... or do we just delete our repos and automatically include dependencies to metacity/panels/applets?

lukefromdc:
Mate-panel also has some custom theme support elements added concerning the GTK theme applied to menus. I use that in my GTK theme (UbuntuStudio_Legacy) to use a panel menu theme ("shell theme") similar to what early versions of gnome-shell used, and some of the calendar work is also used in Blue-Submarine and Green-Submarine. Also we don't require using an obscure key combination to unlock the panel right-click menus. 

To combine the panel projects, I could refactor the panel menu style class (don't use mate-panel-menu-bar but rather something like mate-panel-menu or mate-panel-menu-window) for the theme support so that we would not be combining in a mess with the style class names. I favor moving all the applets in-process as that solves a lot of problems with future Wayland porting, and gnome-panel's newer codebase seems to be a bit less buggy in some ways.  Porting in the refactored theming support, reverting the locked right-click menus,  porting over certain new features we've added such as being able to reset the panel from the context menu, and then combining the projects would in my judgement gives us a panel with exactly the same desired behavior, a little less resource consumption (as everything in in-process), and hopefully fewer bugs.   Even if the projects stayed separate, a "fast-forward" refork of mate-panel from current gnome-panel would have a lot of benefits assuming we also port to in-process any applet we have that GNOME does not (are there any?).

Caja would benefit from a (brutally complex though) refactoring to use the current "backend" code from Nautilus as much as possible while retaining Caja's unchanged UI frontend. For the split-out desktop code the last version of Nautilus-desktop would offer a starting point, though the conversion of the icon rename labels to popovers might need either restyling the popovers (to use the same code) or reverting that commit(by hand if necessary).

Our team is small, and the gnome-flashback team even smaller. Lots of duplicated work and minor differences. The combined project would simply need to handle the use-cases, styles, and developer needs of both and that part should not be a problem

muktupavels:
On Tue, Aug 28, 2018 at 10:49 PM Victor Kareh <notifications@github.com>
wrote:

> So how do we collaborate on this then? We need to settle a few things then:
>
Sure, there is no point in doing anything if there is no good plan. Read
this - https://ometer.com/features.html. If you don't agree to that, then
we already might have a problem. Wrong indentation and/or not following
coding style will be good reason to not accept patch/merge!

> *Window manager*
> I agree that metacity is more stable, so we would need to backport new
> marco features into metacity and make metacity the default window manager.
>
I hope that you did read the link I gave. What features?

> How do we ensure that metacity development follows the mate ideal without
> breaking gnome flashback?
>
Lets be clear - Metacity will not follow MATE ideal or anything like that.
Same is true for GNOME Flashback! Metacity is Metacity and that will not
change. It is not GNOME Flashback window manager, GNOME Flashback use
Metacity as window manager. Maybe might be good idea to list your plans?

For example, Metacity will never be Wayland compositor, it is and will
remain x11 window manager.

> Could we implement features in metacity that @muktupavels
> <https://github.com/muktupavels> is not interested in, but that would
> benefit mate?
>
Just because there might be feature that I am not interested should not be
reason to reject it, but read the link!

> *Panels*
> We would need to settle the in-process vs out-of-process debacle.
>
In process! I definetly don't want to return to GtkSocket/GtkPlug...

> Also, gnome-panel pulls in a lot of gnome dependencies, including nautilus.
>
What dependencies are problem for you? Nautilus is not dependency for
`gnome-panel` , so that is wrong...

> So we would need to retain mate-panel, except for that means deviating
> code-bases. So how do we keep collaborating in panels without breaking
> either mate or gnome flashback? Seems like here we need to keep our own for
> now...
>
As you probably know, then currently panel have applets and internal
"objects". I am slowly moving away from internal objects. So everyting on
panel at some point will be applets. While doing that I am also creating
new library. Few things that are planned:
- Add option for applet to show initial configuration window / dialog. For
example adding weather applet to panel could allow to configure location
before it is added to panel.
- Provide way to fill about dialog. My idea is that there could be "About"
button in Add to panel window, but main idea is that applet itself should
not care where that button and/or menu item will apear.

Since new library is still private, there is time to change API if that is
needed. So feel free to review it:
https://gitlab.gnome.org/GNOME/gnome-panel/tree/master/libgnome-panel

> *Applets*
> We could conceivably merge these into a single repo, or split them all
> into individual repos, I'm fine with either...
>
> If those are really the only components to gnome flashback, then this
> could be maneagable (apart from the dependencies aspect).
>
Again, list with dependencies that are or could be problem.

> There's also the logistics: how do we merge the teams? does everyone want
> this? where do we host stuff? etc. etc... or do we just delete our repos
> and automatically include dependencies to metacity/panels/applets?
>
https://gitlab.gnome.org/GNOME/gnome-panel/
https://gitlab.gnome.org/GNOME/gnome-applets/
https://gitlab.gnome.org/GNOME/metacity/

I don't see any good reason to delete repos! Why would you do that?

Also there is no need to merge teams. Simply instead of creating merge
requests in github you will do that in gitlab.gnome.org. Bugzilla is past,
so workflow now should be similar, no?

On Tue, Aug 28, 2018 at 11:50 PM lukefromdc <notifications@github.com>
wrote:

> Mate-panel also has some custom theme support elements added concerning
> the GTK theme applied to menus. I use that in my GTK theme
> (UbuntuStudio_Legacy) to use a panel menu theme ("shell theme") similar to
> what early versions of gnome-shell used, and some of the calendar work is
> also used in Blue-Submarine and Green-Submarine.
>
?

> Also we don't require using an obscure key combination to unlock the panel
> right-click menus.
>
Well I don't like that too...

> To combine the panel projects, I could refactor the panel menu style class
> (don't use mate-panel-menu-bar but rather something like mate-panel-menu or
> mate-panel-menu-window) for the theme support so that we would not be
> combining in a mess with the style class names.
>
I think you have used too much style classes then really needed.

> I favor moving all the applets in-process as that solves a lot of problems
> with future Wayland porting, and gnome-panel's newer codebase seems to be a
> bit less buggy in some ways.
>
I don't see Wayland "support" in GNOME Flashback anytime soon... And have
not seen any work for that in MATE too.

rilian-la-te:
>- Provide way to fill about dialog. My idea is that there could be "About"
button in Add to panel window, but main idea is that applet itself should
not care where that button and/or menu item will apear.

What do you think about [this way](https://gitlab.com/vala-panel-project/vala-panel/blob/master/ui/applet-info.c) (fill dialog from values from file where plugin name and description appears)

muktupavels:
https://gitlab.gnome.org/GNOME/gnome-panel/blob/master/libgnome-panel/gp-module.c#L47

Idea is that developer provides callback function. If that is set then panel knows that applet has / wants about dialog so it can add button or menu item somewhere. When it is clicked panel creates `GtkAboutDialog` and calls that callback function and after that shows dialog.

muktupavels:
@rilian-la-te I don't want separate file for plugin / module info...

raveit65:
I am not sure if we should start with panel and panel-applets, because we lost menu icons and it isn't clear how we should handle GtkStatusIcon which will gone with gtk+-4.
With using mate-panel/mate-panel-applets we have the option to fork the needed code.

Honestly, i want to keep those features as long as possible, as those are typical elements of the  'look&feel' of Mate which our users expect from us.
Also, it is better to start with a smaller single step to test how a partial fusion can work.

As metacity/marco don't have so much dependency, metacity seems to be the only candidate for integration to Mate for me in the moment.
Here we can test how things works for us without given up our positions.
And i don't see any reason not to use marco as fallback VM.

@monsta @flexiondotorg @XRevan86 @clefebvre 
Please contribute to this discussion

rilian-la-te:
> because we lost menu icons

No, you can use menu-icons actually by this:
```c
void apply_menu_properties(GList *w, GMenuModel *menu)
{
	GList *l = w;
	for (int i = 0; i < g_menu_model_get_n_items(menu); i++)
	{
		uint jumplen = 1;
		if (GTK_IS_SEPARATOR_MENU_ITEM(l->data))
			l = l->next;
		GtkMenuItem *shell            = GTK_MENU_ITEM(l->data);
		const char *str               = NULL;
		bool has_section              = false;
		bool has_submenu              = false;
		GtkMenuShell *menuw           = GTK_MENU_SHELL(gtk_menu_item_get_submenu(shell));
		g_autoptr(GMenuLinkIter) iter = g_menu_model_iterate_item_links(menu, i);
		GMenuModel *link_menu         = NULL;
		while (g_menu_link_iter_get_next(iter, &str, &link_menu))
		{
			bool is_section = !(strcmp(str, G_MENU_LINK_SECTION));
			bool is_submenu = !(strcmp(str, G_MENU_LINK_SUBMENU));
			if (menuw != NULL && is_submenu)
			{
				g_autoptr(GList) children =
				    gtk_container_get_children(GTK_CONTAINER(menuw));
				apply_menu_properties(children, link_menu);
			}
			if (is_section)
			{
				jumplen += ((uint)g_menu_model_get_n_items(link_menu) - 1);
				apply_menu_properties(l, link_menu);
			}
			g_object_unref(link_menu);
			has_section = has_section || is_section;
			has_submenu = has_submenu || is_submenu;
		}
		GVariant *val = NULL;
		g_autoptr(GMenuAttributeIter) attr_iter =
		    g_menu_model_iterate_item_attributes(menu, i);
		while (g_menu_attribute_iter_get_next(attr_iter, &str, &val))
		{
			if (!strcmp(str, G_MENU_ATTRIBUTE_ICON) && (has_submenu || has_section))
			{
				g_autoptr(GIcon) icon = g_icon_deserialize(val);
				g_object_set(shell, "icon", icon, NULL);
			}
			if (!strcmp(str, ATTRIBUTE_TOOLTIP))
				gtk_widget_set_tooltip_text(GTK_WIDGET(shell),
				                            g_variant_get_string(val, NULL));
			if (!strcmp(str, ATTRIBUTE_DND_SOURCE) && g_variant_get_boolean(val))
				apply_menu_dnd(GTK_MENU_ITEM(l->data), menu, i);
			g_variant_unref(val);
		}
		l       = g_list_nth(l, jumplen);
		jumplen = 1;
		if (l == NULL)
			break;
	}
}
```

This code actually appends menu icons, tooltips and DnD sources to MenuModel created menus.

rilian-la-te:
@muktupavels you do not need. You can use [this type of file](https://gitlab.com/vala-panel-project/vala-panel-appmenu/blob/master/data/appmenu-mate.desktop.plugin.in)
(this is generator approach to applet factory file, you can add about info into it)

muktupavels:
@rilian-la-te In `gnome-panel` case there would be only `libappmenu-mate.so`...

Anyway this pull request is not right place for this discussion, is it? If you are interested then probably might be good idea to open new issues for each module (marco, mate-panel?) and continue discussion there...

rilian-la-te:
@muktupavels this is for MATE panel)

lukefromdc:
BTW, I found this breaks builds of compiz-reloaded, as gtk-window-decorator built against older Marco segfaults when run against this. Rebuilding compiz 0.8 against current Marco fixes the problem. Anyone using or distributing compiz-reloaded will need to be aware of this

raveit65:
@vkareh 
Huston, we got a problem.
m-c-c doesn't build any more!
```
theme-thumbnail.c:4:10: fatal error: marco-private/util.h: No such file or directory
 #include <marco-private/util.h>
          ^~~~~~~~~~~~~~~~~~~~~~
```

raveit65:
I guess build error with m-c-c is caused by https://github.com/mate-desktop/marco/pull/422/commits/940e615ef9487c39295d10b0239c7ee968eb0bce

rilian-la-te:
> Bad idea......, the idea was to use metacity
@raveit65 confirmed) For me gnome-panel is just ... weird. I am sorry.
I do not like mate-panel too, but I think mate-panel is superior.

PR #424, opened 2018-08-29
backport: metacity-theme-3 (part 2)

vkareh:
This bumps metacity-theme-3 to version 3.6, adding invisible borders and app-menu buttons. Neither are displayed, but the theme recognizes them.

Also reverts commit 2bea67fce88dc22320874279faf08da72ba02ca5 that breaks m-c-c build


muktupavels:
https://gitlab.gnome.org/GNOME/metacity/commit/37a2d2056f05c940c6244627da25057f42be8807

vkareh:
@muktupavels - why was it removed, I'm confused? the commit message doesn't say anything, it would be helpful if there was a reason stated for that.

muktupavels:
@vkareh Unfortunately I am very, very bad at writing good commit messages...

Reason is - that was never part of metacity-theme-3... That format was created by mutter, it is simply bad idea to introduce incompatible changes. What would happen if mutter would add something else with that version? Of course mutter no longer use it, but if it was not there then it should not be anywhere else.

You already have change that is incompatible, and `metacity` also had it at some point:
https://github.com/mate-desktop/marco/pull/422/commits/2b2e45b0abb2239e01d1b442a8665804aa5a8b78

Shadow and padding was never part of theme, it is always ubuntu specific change. Add that to theme and it will crash `metacity` or other forks. And it is theme bug / problem because it is using something that is not available...

If you want something that is not in metacity (format 1 & 2) and mutter (format 3) then only correct thing is to move to macro specific format, like macro-theme-4.xml.

vkareh:
Good, thanks for the explanation. I will revert that commit.

> the only correct thing is to move to marco specific format, like marco-theme-4.xml

All in due time, first I want to bring marco up to full compat with metacity-theme-3.xml, then we can start talking about moving forward from there. :smile:

lukefromdc:
Not surprisingly, I had to rebuild compiz-core (compiz-reloaded/0.8.15 version) again with the latest changes. No problem so long as no rolling distro builds compiz 0.8 against Marco 1.20 and then updates Marco to 1.22 without rebuilding compiz. No source code had to change though.

lukefromdc:
I also just confirmed that mate-control-center now configures and builds just fine with this, and an old build (from before these PR's) ran just fine, including the windows capplet. Toggling transparency as a test worked, showing that an existing build of mate-control-center can still connect to Marco after these changes

muktupavels:
- https://github.com/mate-desktop/marco/blob/master/src/Makefile.am#L128
- https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

lukefromdc:
I don't know anything about libtool version stuff. Are there problems here that need to be fixed in a new PR?

PR #439, opened 2018-12-31
compositor-xrender: avoid deprecated 'gdk_error_trap_push/pop'

sc0w:


muktupavels:
Why not `meta_error_trap_push` and `meta_error_trap_pop_with_return`?

sc0w:
@muktupavels 

because that functions have deprecated 'gdk_error_trap_push' and deprecated 'gdk_error_trap_pop'

https://github.com/mate-desktop/marco/blob/master/src/core/errors.c#L34

https://github.com/mate-desktop/marco/blob/master/src/core/errors.c#L53

muktupavels:
Why would that be problem?

sc0w:
In the fixes of deprecated functions I like to respect the original style, the deprecated function is here and I fixed it here

muktupavels:
I understand, but why do this to redo that later? That is only place where gdk functions are used directly if we don't count `errors.c`. Also it seems that it is only reason why `gtk/gtk.h` and `gdk/gdk.h` is included here...

raveit65:
> Why not `meta_error_trap_push` and `meta_error_trap_pop_with_return`?

Why?
I see no reason to replace a deprecated function with a function which has deprecated elements.

muktupavels:
Do you know why these functions was deprecated? Do you know why these functions are needed?

I am just suggesting change that would be better in my opinion. Is not that point of reviews? Spotting mistakes, suggesting better solutions? Otherwise what is point of review for this? Change is simple enough and could have been directly pushed to master...

All codebase use `meta_error_trap_*` functions, I would say that these gdk functions should never been added there in first place (that was done already in metacity).

Doing that code would follow same style and you will be able to remove now unneeded includes. Not my suggestion nor change in this merge request change functionality...

raveit65:
Please do a PR if you know it better than all other people here ;)

PR #442, opened 2019-01-06
add git.mk to autogenerate .gitignore files

yetist:
When running `make`, autogenerate the `.gitignore` files to keep the clean output of `git status`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,3 +17,5 @@ ChangeLog:
 dist: ChangeLog
 
 .PHONY: ChangeLog
+
+-include $(top_srcdir)/git.mk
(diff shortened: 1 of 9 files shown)


lukefromdc:
I don't know enough about this to properly evaluate a test

raveit65:
This works and generated make files doesn't change git status.
But keep in mind that before you have to run `make` you need to execute `./autogen.sh` which generates some more files.
```
[rave@mother marco-dist]$ git status
On branch add_gitmk
Your branch is up to date with 'origin/add_gitmk'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	INSTALL
	aclocal.m4
	compile
	config.guess
	config.h.in
	config.sub
	depcomp
	install-sh
	ltmain.sh
	m4/
	missing
	omf.make
	xmldocs.make

nothing added to commit but untracked files present (use "git add" to track)
```
So, personal i wouldn't run `make` in same repo where i push to github.

PR #443, opened 2019-01-08
prefs: avoid deprecated 'g_settings_list_keys'

sc0w:


lukefromdc:
Segfault is
https://github.com/mate-desktop/marco/issues/445
and appears only when window thumbnails are displayed in the alt-tab switcher with compositing enabled. Has nothing whatsoever to do with this PR as changing the preferences (including to test this) works fine either way. 

Thus this should be safe to merge as it changes only reading and writing of settings, I tested changing workspace names in particular as that is mentioned in the code so was easily tested and worked perfectly. Three approving reviews so merging

lukefromdc:
Rebased on latest master

PR #447, opened 2019-01-22
Increase icon size on tab and workspace popups

### LINKED ISSUE #333: Alt+tab: icons are too small
#### Actual behaviour
I use Marco with enabled compositor and with disabled windows thumbnails. When I press Alt+tab to switch windows, I see windows icons — they are too small and it is hard to recognize applications.

#### Expected behaviour
Windows icons should be bigger, such as in Metacity.

#### Package version
marco 1.18.1

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47267316-alt-tab-icons-are-too-small?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #363: Why marco switch-tab icon is so small when composite-manager is disabled
#### Expected behaviour
Make switch-tab icon larger for non composite-manager mode, just like what metacity does.

#### Actual behaviour

The icon is too small to see.

Using marco as the window manager, and open multiple window, then press "Alt+Tab".

#### MATE general version

all

#### Package version

all

#### Linux Distribution

ubuntu

vkareh:
Alt+Tab and Workspace popups should be sized relative to the screen size.
This way they look nice and large regardless of the display resolution.

Also, given much larger modern resolutions, icon sizes should be larger by default.

Fixes #333 
Fixes #363

### DIFF
--- a/src/core/screen-private.h
+++ b/src/core/screen-private.h
@@ -36,7 +36,6 @@
 #include "display-private.h"
 #include "screen.h"
 #include <X11/Xutil.h>
-#include <gdk/gdk.h>
 #include "ui.h"
 
 typedef struct _MetaXineramaScreenInfo MetaXineramaScreenInfo;
@@ -193,8 +192,6 @@ void          meta_screen_get_natural_xinerama_list (MetaScreen *screen,
 void          meta_screen_update_workspace_layout (MetaScreen             *screen);
 void          meta_screen_update_workspace_names  (MetaScreen             *screen);
 void          meta_screen_queue_workarea_recalc   (MetaScreen             *screen);
-
-GdkMonitor* meta_screen_get_current_monitor (void);
 
 Window meta_create_offscreen_window (Display *xdisplay,
                                      Window   parent,
--- a/src/core/screen-private.h
+++ b/src/core/screen-private.h
@@ -36,6 +36,7 @@
 #include "display-private.h"
 #include "screen.h"
 #include <X11/Xutil.h>
+#include <gdk/gdk.h>
 #include "ui.h"
 
 typedef struct _MetaXineramaScreenInfo MetaXineramaScreenInfo;
@@ -192,6 +193,8 @@ void          meta_screen_get_natural_xinerama_list (MetaScreen *screen,
 void          meta_screen_update_workspace_layout (MetaScreen             *screen);
 void          meta_screen_update_workspace_names  (MetaScreen             *screen);
 void          meta_screen_queue_workarea_recalc   (MetaScreen             *screen);
+
+GdkMonitor* meta_screen_get_current_monitor (void);
 
 Window meta_create_offscreen_window (Display *xdisplay,
                                      Window   parent,
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -1224,8 +1224,7 @@ get_window_pixbuf (MetaWindow *window,
   MetaDisplay *display;
   cairo_surface_t *surface;
   GdkPixbuf *pixbuf, *scaled;
-  GdkMonitor *monitor;
-  GdkRectangle rect;
+  const MetaXineramaScreenInfo *current;
   double ratio;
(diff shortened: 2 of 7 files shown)


src/core/screen.c
```diff
@@ -1246,20 +1247,25 @@ get_window_pixbuf (MetaWindow *window,
   *width = gdk_pixbuf_get_width (pixbuf);
   *height = gdk_pixbuf_get_height (pixbuf);
 
-  /* Scale pixbuf to max dimension MAX_PREVIEW_SIZE */
+  /* Scale pixbuf to max dimension based on screen sizscreen size */
```
muktupavels: I guess you are ignoring multi-monitor setups... For example if I have 3 1280x1024 monitors side by side I would have 3840px screen width. That would be 384px as preview width, no?

vkareh: @muktupavels - yes, good point. I'm working on a fix for this.

vkareh: okay, I changed it to be based on monitor size, rather than screen size

vkareh:
@lukefromdc - yeah, I can see the huge icons in the workspace switcher in _some_ instances. I think making the switcher itself smaller will scale down the icons, so I will try that.

raveit65:
ui/tabpopup.c has some deprecated GtkMisc warnings, but they are in master already.

vkareh:
@raveit65 - fixed the function prototype (was missing the `void` inside the prototype parameters)

src/core/screen-private.h
```diff
@@ -193,6 +194,8 @@ void          meta_screen_update_workspace_layout (MetaScreen             *scree
 void          meta_screen_update_workspace_names  (MetaScreen             *screen);
 void          meta_screen_queue_workarea_recalc   (MetaScreen             *screen);
 
+GdkMonitor* meta_screen_get_current_monitor (void);
```
muktupavels: Why this is put here?  Can not it be put in somewhere in `src/ui`?

vkareh: What would be the difference? (honest question, I actually struggled trying to figure out where to put it)

muktupavels: I simply think that it is best to limit code availability to place where it is needed. Why put in `screen.c` if it used only in ui? Otherwise there is no functional difference.

muktupavels:
Also is it really needed to go through gdk? You should be able to get this info from `marco` itself...

muktupavels:
@vkareh yes.

vkareh:
@muktupavels - also, why is going through gdk a problem? If I use xinerama, I'm doing a roundtrip to the X server anyway, which is the same that gdk uses internally. Is the problem that GDK would get all this information from scratch, while marco already has it in memory? It almost seems as though I could replace the `XQueryPointer` code with the `GdkDevice`, and not have to maintain that calculation inside marco, but I'm not entirely sure which one is more efficient

vkareh:
There, I pushed a new commit that removes the `meta_screen_get_current_monitor` function and reuses the old `meta_screen_get_current_xinerama`. I kept it as a separate commit in case we don't want it, to not mess around with the previous working commit.

@muktupavels - thoughts?

vkareh:
@muktupavels - cool, thanks!

@raveit65 - are you ok with the new version of the code? I removed the new function and re-used the existing xinerama one.

raveit65:
yeep, i am very happy not to test it again :)

lukefromdc:
Just retested one more time. Good enough, and on small but highdpi screens offers a usability advantage

rolandua:
Is it able to change Alt-Tab previews size somewhere in gtk.css or dconf?

PR #449, opened 2019-02-03
Enable Travis CI

yetist:
It is ready to go.


raveit65:
All checks are successful ;)
I will squash both commits before merging.

PR #455, opened 2019-02-18
frames: avoid deprecated 'gtk_widget_set_double_buffered'

sc0w:
please, don't push to the 1.22 release

EDIT: latest changes:

based in the mutter commits:

https://gitlab.gnome.org/GNOME/mutter/commit/b8b0dc5c78008647f5833f63d88456025f17bd08

https://gitlab.gnome.org/GNOME/mutter/commit/669c9da2a48d9a5573edee09065b8558c1d07035

### DIFF
--- a/src/ui/frames.c
+++ b/src/ui/frames.c
@@ -198,8 +198,6 @@ meta_frames_init (MetaFrames *frames)
   frames->invalidate_frames = NULL;
   frames->cache = g_hash_table_new (g_direct_hash, g_direct_equal);
 
-  gtk_widget_set_double_buffered (GTK_WIDGET (frames), FALSE);
-
   meta_prefs_add_listener (prefs_changed_callback, frames);
 }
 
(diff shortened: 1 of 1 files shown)


sc0w:
@lukefromdc yes, confirmed.

With latest changes: I have tested and all seems to work fine, the build is ok without no new warnings in the affected code, and clang static analyzer doesn't show me new warnings.

@lukefromdc please test again.

raveit65:
Is this really save to remove?
I found only this discussions which doesn't make it more clearer.
https://mail.gnome.org/archives/gtk-devel-list/2014-November/msg00039.html
If i understand it correctly, than it seems only save to remove for wayland, but not for x11.
But maybe you found better docs than me?

raveit65:
Here is a good explanation for this function.
https://people.gnome.org/~shaunm/girdoc/C/Gtk.Widget.set_double_buffered.html

sc0w:
@raveit65 

It is needed remove it, in gtk4 the function is gone:

https://gitlab.gnome.org/GNOME/gtk/commit/4a95f8802c22971976dc50e8818af6256218a54d

The commits from mutter here have 4 years old without changes, so, I suppose this code is safe.

PR #456, opened 2019-02-22
Enable Clang Static Analyzer for marco

yetist:
Let me try, if every thing is ok, the gh-pages branch will have many html files.


yetist:
PR does not trigger deployment.

PR #458, opened 2019-02-25
compositor-xrender: Fix memory leak

sc0w:
Fixes Clang static analyzer warning:

```
compositor/compositor-xrender.c:587:10: warning: Potential leak of memory pointed to by 'data'
  return ximage;
         ^~~~~~
```

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -584,6 +584,8 @@ make_shadow (MetaDisplay   *display,
         }
     }
 
+  ximage->data = (char *) data;
+
   return ximage;
 }
 


src/compositor/compositor-xrender.c
```diff
@@ -584,6 +584,8 @@ make_shadow (MetaDisplay   *display,
         }
     }
 
+  g_free (data);
```
rbuj: You can't do `g_free (data);` here, **ximage->data** points to **data**. 
https://tronche.com/gui/x/xlib/utilities/XCreateImage.html

sc0w: ok, this new patch works:

```patch
diff --git a/src/compositor/compositor-xrender.c b/src/compositor/compositor-xrender.c
index c18c8dd..de112db 100644
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -584,6 +584,8 @@ make_shadow (MetaDisplay   *display,
         }
     }
 
+  ximage->data = (char *) data;
+
   return ximage;
 }
```

@rbuj  are you agree?

rbuj: yes

muktupavels: @sc0w what `ximage->data = (char *) data;` is supposed to do?

sc0w: @rbuj ok, done!

@muktupavels it fixes the warning I mentioned

muktupavels: Then perhaps add comment explaining need for that? Because otherwise it is useless code... It does nothing, no? https://gitlab.freedesktop.org/xorg/lib/libx11/blob/master/src/ImUtil.c#L360

muktupavels:
This is false positive, no? Warnings itself says "Potential leak"... Looking at documentation and source code of `XCreateImage` and `XDestroyImage` you can see that there is no leak. Or did I miss something?

Adding redundant line of code to shut up warning does not look like correct thing to me... As minimum add comment explaining that it is used only to remove warning. But there might be better way to suppress this warning.

PR #459, opened 2019-02-26
screen: Fix use of memory after it is freed

sc0w:
Fixes Clang static analyzer warning:

```
core/screen.c:754:16: warning: Use of memory after it is freed
      result = g_list_prepend (result, info);
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -749,9 +749,8 @@ list_windows (MetaScreen *screen)
       else
         {
 	  info->xwindow = children[i];
+          result = g_list_prepend (result, info);
 	}
-
-      result = g_list_prepend (result, info);
     }
 
   if (children)


muktupavels:
This does not look correct to me... Now you have windows list with data pointers to freed memory, no? I think this might be missing `continue;` after free. Or you could add `NULL` check for `g_list_prepend` and reset `info` to `NULL` after freeing it in error case.

sc0w:
@muktupavels do you think this change is better? (patch over git master)

```patch
diff --git a/src/core/screen.c b/src/core/screen.c
index c03530c..e0e4e50 100644
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -749,9 +749,8 @@ list_windows (MetaScreen *screen)
       else
         {
 	  info->xwindow = children[i];
+          result = g_list_prepend (result, info);
 	}
-
-      result = g_list_prepend (result, info);
     }
 
   if (children)
```

muktupavels:
That should work.

sc0w:
@muktupavels thanks for review, I will do it later

src/core/screen.c
```diff
@@ -749,9 +749,8 @@ list_windows (MetaScreen *screen)
       else
         {
 	  info->xwindow = children[i];
+          result = g_list_prepend (result, info);
```
muktupavels: Indentation...

sc0w: seems to be ok here:



I use spaces, in the code in some lines have [spaces], in others [tab+spaces], it seems to be randomly in the function

muktupavels: You are right, above line is wrong, but that is not related to your change. :) Might be good idea to fix that and remove tabs in separate pull request.

PR #461, opened 2019-02-27
Improve performance by removing unused assignments

rbuj:


src/core/boxes.c
```diff
@@ -577,7 +577,6 @@ meta_rectangle_get_minimal_spanning_set_for_region (
   *temp_rect = *basic_rect;
   ret = g_list_prepend (NULL, temp_rect);
 
-  strut_iter = all_struts;
```
muktupavels: This most likely is optimized out by compiler... So I doubt it improves anything.

src/core/edge-resistance.c
```diff
@@ -285,7 +285,6 @@ find_nearest_position (const GArray        *edges,
           if (dist < best_dist)
             {
               best = compare;
-              best_dist = dist;
```
muktupavels: Care to explain how this is unused?

src/core/window.c
```diff
@@ -3597,11 +3597,6 @@ meta_window_move_resize_internal (MetaWindow          *window,
                   window->frame->rect.width,
                   window->frame->rect.height);
     }
-  else
-    {
-      frame_size_dx = 0;
```
muktupavels: How about reducing scope by moving `frame_size_dx` and `frame_size_dy` into if branch if they are used only there?

src/ui/frames.c
```diff
@@ -1492,8 +1492,6 @@ meta_frames_button_press_event (GtkWidget      *widget,
           op = op > META_GRAB_OP_CLICKING_MAXIMIZE_HORIZONTAL ? META_GRAB_OP_CLICKING_MAXIMIZE : op;
           break;
         case META_FRAME_CONTROL_UNMAXIMIZE:
-          op = META_GRAB_OP_CLICKING_UNMAXIMIZE + event->button - 1;
-          op = op > META_GRAB_OP_CLICKING_UNMAXIMIZE_HORIZONTAL ? META_GRAB_OP_CLICKING_UNMAXIMIZE : op;
```
muktupavels: Can you explain? Looking at `META_FRAME_CONTROL_MAXIMIZE` I would say that `op = META_GRAB_OP_CLICKING_UNMAXIMIZE;` is the line that should be removed... Might be good idea to check history of this code.

src/ui/frames.c
```diff
@@ -1525,7 +1523,7 @@ meta_frames_button_press_event (GtkWidget      *widget,
           break;
         default:
           g_assert_not_reached ();
-          break;
```
muktupavels: Why are you removing `break`?

src/core/edge-resistance.c
rbuj: there is a break below

src/ui/frames.c
rbuj: op is overwritten at the end of the case

src/ui/frames.c
rbuj: default don't require break

src/core/edge-resistance.c
muktupavels: Ah, I was looking at wrong line... Sorry!

src/ui/frames.c
muktupavels: Yes, but are you sure that is right thing? Perhaps you just found source for some bug...

src/core/window.c
rbuj: great idea!

src/ui/frames.c
muktupavels: It does not, but it is not reason to remove it, is it? I am with those who think you should always add `break` - https://softwareengineering.stackexchange.com/questions/201777/break-on-default-case-in-switch

src/core/window.c
rbuj: done

muktupavels:
Since title implies that this improves performance where are numbers before and after?

Mose if not all changes might have been already optimized by compiler... I would prefer more readable / safer code. For example initializing variable before switch. What if someone in future add new case that does not initialize variable? Sure that would be programmers problem/bug/error, but still...

src/core/edge-resistance.c
rbuj: don't worry

src/core/boxes.c
rbuj: source code can not depend on binary optimization done by the compiler for unfound sentences like this.

muktupavels: I am not saying that you should relay that compiler will optimize it out. Of course this is obviously unneeded assignment and can be removed.

You are claiming to improve performance, where are numbers? And under what test conditions?

Any sane compiler will optimize out this redundant assignment so unless you disable optimizations this improves performance by exactly 0%. `-01` with gcc will be enough to optimize out redundant assignments or completely remove unused variables. So in any real life usage there wont be any performance improvement.

And even if optimizations are disabled what are you gaining? Save 1 cpu cycle (of course with modern cpus that is more complicated)?

PR #462, opened 2019-02-28
MATE uses header bars windows problematic

zhuyaliang:
I don't understand why we should judge whether the window is decorated or not.I think it would be better to remove this judgment so that problems with header-bar windows can be avoided.

### DIFF
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -1528,7 +1528,6 @@ constrain_titlebar_visible (MetaWindow         *window,
       window->type == META_WINDOW_DOCK    ||
       window->fullscreen                  ||
       !window->require_titlebar_visible   ||
-      !window->decorated                  ||
       unconstrained_user_action)
     return TRUE;
 


raveit65:
I like this change because it fixes a serious problem with CSD windows.
Without this PR it is possible to move a CSD window under to top panel by accidental, and you can't reach the close/minimize buttons of the headerbar any more.
With the PR this is fixed for the case that one monitor is used.

With a multi monitor-setup i can move the CSD window only on one side of the monitor under the top panel.
See this video https://www.dropbox.com/s/0mmzdz3g6x9dh6m/csd-panel.ogv?dl=0
(ignore quarter-tiling issue of second monitor, not related to this issue)
Any way, i think panel struts a quiet broken for multi-montior setups, so this problem isn't releated to PR.

So this is a great improvement for handling CSD windows.

lukefromdc:
I can see the issue this is aimed  right away with gnome-disks: you can slide the header bar under a top panel with the mouse

PR #464, opened 2019-03-02
delete: remove old code

rbuj:
See https://gitlab.gnome.org/GNOME/metacity/commit/da6ff49008e3a20a4bac05e10a0ac62035f472ae


muktupavels:
Include original author and date.

rbuj:
@muktupavels 
done

muktupavels:
That is not what I meant... Why you just don't use original commit?

rbuj:
Imported in PR #466

PR #473, opened 2019-03-08
testboxes: Fix find_closest_point_to_line() test

rbuj:
check whether two floating point numbers are equal within
an epsilon

### DIFF
--- a/src/core/boxes.c
+++ b/src/core/boxes.c
@@ -29,6 +29,9 @@
 #include "boxes.h"
 #include "util.h"
 #include <X11/Xutil.h>  /* Just for the definition of the various gravities */
+#include <math.h>
+
+#define POLINOMIAL_WITH_SINGLE_NEGATIVE_ROOT(a, b) pow(a,2) - 2*a*b + pow(b,2)
 
 char*
 meta_rectangle_to_string (const MetaRectangle *rect,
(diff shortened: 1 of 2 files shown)


muktupavels:
How about including some info about what is actual problem you are trying to fix?

rbuj:
We can find the closest point to the diagonal of the rectangle, instead of the rectangle itself, which can produce an overflow.
Once we have the point on the diagonal, if exists, we have to find the closest point for the four lines of the rectangle, the line will have the shortest distance. Finally, we have to find the point of the line.

muktupavels:
Why are you changing test data? Wrong?

It still not clear to me what is actual problem with code... Do you have example how this functions fails / causes problems under real usage?

rbuj:
The new examples were tested with numpy and shapely, because I don't know where previous ones come from (Mathematica, Mapple...), however they were correct.
```sh
sudo dnf install -y python3-shapely python3-numpy
```
Tests:
Test I: vertical line
```python
import numpy as np
from shapely.geometry import Point
from shapely.geometry import LineString
line = LineString([(3, 49), (3, -1)])
point = Point(19.1, -2.6)
x = np.array(point.coords[0])
u = np.array(line.coords[0])
v = np.array(line.coords[len(line.coords)-1])
n = v - u
n /= np.linalg.norm(n, 2)
P = u + n*np.dot(x - u, n)
print(P)
```
[ 3.  -2.6]


Test II: horizontal line
Swap the coordinates of the previous example:
```python
import numpy as np
from shapely.geometry import Point
from shapely.geometry import LineString
line = LineString([(49, 3), (-1, 3)])
point = Point(-2.6, 19.1)
x = np.array(point.coords[0])
u = np.array(line.coords[0])
v = np.array(line.coords[len(line.coords)-1])
n = v - u
n /= np.linalg.norm(n, 2)
P = u + n*np.dot(x - u, n)
print(P)
```
[-2.6  3. ]


Test III: oblique line, 45º
```python
import numpy as np
from shapely.geometry import Point
from shapely.geometry import LineString
line = LineString([(0, 0), (50, 50)])
point = Point(25, 25)
x = np.array(point.coords[0])
u = np.array(line.coords[0])
v = np.array(line.coords[len(line.coords)-1])
n = v - u
n /= np.linalg.norm(n, 2)
P = u + n*np.dot(x - u, n)
print(P)
```
[25. 25.]


I didn't reduce the calculus because the numbers are double type, and it could lose precision.

The previous code produces segfault. `./testboxes`

And wm-tester tests work fine, including test-gravity.

muktupavels:
Because assert fails? https://gitlab.gnome.org/GNOME/mutter/commit/0fccb0fc86d4b3054e50aed2ba900d99cb69d48d

rbuj:
@muktupavels 
I go to sleep, I'll review it tomorrow.

lukefromdc:
I don't know enough about this one to properly evaluate it

src/core/boxes.c
```diff
@@ -1122,15 +1123,10 @@ meta_rectangle_find_linepoint_closest_to_point (double x1,
...
-  den = diffx * diffx + diffy * diffy;
-
-  *valx = (py * diffx * diffy + px * diffx * diffx +
-           y2 * x1 * diffy - y1 * x2 * diffy) / den;
-  *valy = (px * diffx * diffy + py * diffy * diffy +
-           x2 * y1 * diffx - x1 * y2 * diffx) / den;
+   *valx = (px*pow(x1,2) - 2*px*x1*x2 + px*pow(x2,2) + x2*pow(y1,2) + x1*pow(y2,2) + (py*x1 - py*x2)*y1 - (py*x1 - py*x2 + (x1 + x2)*y1)*y2) /
+           (pow(x1,2) - 2*x1*x2 + pow(x2,2) + pow(y1,2) - 2*y1*y2 + pow(y2,2));
+   *valy = (py*pow(y1,2) + py*pow(y2,2) + (px*x1 - (px + x1)*x2 + pow(x2,2))*y1 - (px*x1 - pow(x1,2) - (px - x1)*x2 + 2*py*y1)*y2) /
+           (pow(x1,2) - 2*x1*x2 + pow(x2,2) + pow(y1,2) - 2*y1*y2 + pow(y2,2));
```
muktupavels: This is unreadable...

rbuj: I can't decompose it into factors or temporary variables without adding a precision loss with floating-point numbers, like previous code did, probably by forcing base changes o whatever.

rbuj: Moreover, I don't know where the round-off error described in comment comes from, as it doesn't make reference to float.h

muktupavels: In the end result is used as int - https://github.com/mate-desktop/marco/blob/38cd6190c8b2c1553fa814b23734439e0e95f577/src/core/constraints.c#L1343-L1350.

So I doubt precision loss is problem here...

rbuj: Yes, It performs truncation, not rounding, and it should be fixed, for instance 0.9999.... = 0.
I will do another PR fixing that.

rbuj: Use rounded values instead of truncate them https://github.com/mate-desktop/marco/pull/476

raveit65:
Can we close this?
@vkareh 
What do you think?

vkareh:
There is no description regarding what the issue is or how this fixes it. There are no reproduction/test steps. I think we should close. @rbuj any comments/explanations before we close this?

rbuj:
This PR fixes the unit test **test_find_closest_point_to_line**:

Before:
```shell
$ ./testboxes 
test_area passed.
test_intersect passed.
test_equal passed.
test_overlap_funcs passed.
test_basic_fitting passed.
The next test intentionally causes a warning, but it can be ignored.
Window manager warning: Region to merge was empty!  Either you have a some pathological STRUT list or there's a bug somewhere!
test_regions_okay passed.
test_region_fitting passed.
The next test intentionally causes a warning, but it can be ignored.
Window manager warning: No rect whose size to clamp to found!
The next test intentionally causes a warning, but it can be ignored.
Window manager warning: No rect whose size to clamp to found!
test_clamping_to_region passed.
test_clipping_to_region passed.
test_shoving_into_region passed.
test_find_onscreen_edges passed.
test_find_nonintersected_xinerama_edges passed.
test_gravity_resize passed.
**
marco:ERROR:core/testboxes.c:1357:test_find_closest_point_to_line: assertion failed: (rx == answer_x && ry == answer_y)
Bail out! marco:ERROR:core/testboxes.c:1357:test_find_closest_point_to_line: assertion failed: (rx == answer_x && ry == answer_y)
Avortat (s'ha bolcat la memòria)
```
After:
```shell
$ ./testboxes 
test_area passed.
test_intersect passed.
test_equal passed.
test_overlap_funcs passed.
test_basic_fitting passed.
The next test intentionally causes a warning, but it can be ignored.
Window manager warning: Region to merge was empty!  Either you have a some pathological STRUT list or there's a bug somewhere!
test_regions_okay passed.
test_region_fitting passed.
The next test intentionally causes a warning, but it can be ignored.
Window manager warning: No rect whose size to clamp to found!
The next test intentionally causes a warning, but it can be ignored.
Window manager warning: No rect whose size to clamp to found!
test_clamping_to_region passed.
test_clipping_to_region passed.
test_shoving_into_region passed.
test_find_onscreen_edges passed.
test_find_nonintersected_xinerama_edges passed.
test_gravity_resize passed.
test_find_closest_point_to_line passed.
All tests passed.
```

vkareh:
I've created https://github.com/mate-desktop/marco/pull/554 to address this

PR #475, opened 2019-03-10
window: avoid clang warning: garbage value

sc0w:
avoid Clang static analyzer warning:

```
core/window.c:3580:34: warning: The right operand of '+' is a garbage value
      new_w = window->rect.width + fgeom.left_width + fgeom.right_width;
                                 ^ ~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -3484,6 +3484,7 @@ meta_window_move_resize_internal (MetaWindow          *window,
   gboolean is_user_action;
   gboolean configure_frame_first;
   gboolean use_static_gravity;
+  gboolean have_window_frame;
   /* used for the configure request, but may not be final
    * destination due to StaticGravity etc.
    */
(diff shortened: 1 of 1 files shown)


sc0w:
The `gboolean` variable only keeps in RAM while the function is running, and disappears later.

I don't know if this is false positive, but I think `gboolean` is faster than `window->frame` so, I think this PR have better performace

lukefromdc:
OK, that's good enough for me.

muktupavels:
Well, it is false positive, no? Or do you see how this variable can change while this function is running? Ok, function call can be interrupted by hardware interrupt, but marco does not install/use them. Other option would be to change variable under gdb...

Faster? Can you measure it? If not I doubt it matters...

Anyway I would say there is nothing wrong with this change, but I would not make this change just to shut up false positive. Maybe it is worth to report false positives to clang developers?

Actually, why not simply initialize variable?

sc0w:
@muktupavels 

> Maybe it is worth to report false positives to clang developers?

https://clang-analyzer.llvm.org/faq.html

> Actually, why not simply initialize variable?

what do you mean? this change?

`have_window_frame = (window->frame) ? TRUE : FALSE;`

muktupavels:
No, `memset (&fgeom, 0, sizeof (MetaFrameGeometry));`.

PR #476, opened 2019-03-11
Use rounded values instead of truncate them

rbuj:
If the decimal value is from .1 to .5, round(arg) returns
the integer value less than the argument. Otherwise, if
the decimal value is from .6 to .9, it returns the
integer value greater than the arg.

### DIFF
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -23,6 +23,8 @@
  * 02110-1301, USA.
  */
 
+#define _GNU_SOURCE
+
 #include <config.h>
 #include "constraints.h"
 #include "workspace.h"
@@ -1327,9 +1329,8 @@ constrain_aspect_ratio (MetaWindow         *window,
                                                       new_width, new_height,
                                                       &best_width, &best_height);
 
-      /* Yeah, I suck for doing implicit rounding -- sue me */
-      new_width  = best_width;
-      new_height = best_height;
+      new_width  = round (best_width);
+      new_height = round (best_height);
 
       break;
     }
--- a/configure.ac
+++ b/configure.ac
@@ -29,6 +29,8 @@ m4_define([lt_version_info], [lt_current:lt_revision:lt_age])
 
 AC_SUBST([LT_VERSION_INFO], [lt_version_info])
 
+AC_CHECK_LIB(m, round)
+
 # Honor aclocal flags
 AC_SUBST(ACLOCAL_AMFLAGS, "\${ACLOCAL_FLAGS}")
 


sc0w:
travis says:

```
configure: error: Could not find a C99 compatible compiler
make: *** [Makefile:677: distcheck] Error 1
!!! ERROR: run command [docker exec -t marco-debian-build /rootdir/after_scripts].
The command "./docker-build --name ${DISTRO} --verbose --config .travis.yml --build scripts" exited with 1.
```

lukefromdc:
Is this something we need to fix and merge or can it be closed? I do not know what the functional effect is supposed to be. I see it fails the Travis builds, but over some kind of compiler (C99) issue

raveit65:
Can we close this?

rbuj:
The configure error is fixed.

src/core/constraints.c
```diff
@@ -1345,9 +1347,8 @@ constrain_aspect_ratio (MetaWindow         *window,
                                                       new_width, new_height,
                                                       &best_width, &best_height);
 
-      /* Yeah, I suck for doing implicit rounding -- sue me */
-      new_width  = best_width;
-      new_height = best_height;
+      new_width  = round (best_width);
+      new_height = round (best_height);
```
cwendling: As it's cast to int anyway, why not use this simple version that avoids using the C99 `round()` that introduce a new check that even failed at some point?
```c
      new_width  = (int) (best_width + 0.5);
      new_height = (int) (best_height + 0.5);
```
The computation result is the same (or at least virtually so, I didn't check the standard to see if that can result in a slightly different rounding method, but it should be and a program built with GCC gives exact same results).

cwendling:
This said, is this a fix or something?  There don't seem to be an explanation why this piece of code would have had a problem with the previous code (yes, it says "rounding", but that could merely meant "make an integer out of it" without the specific meaning of "nearest integer").
Changing this could result in wrong/suboptimal results if it wasn't an actual issue.

raveit65:
I am testing.....
@rbuj 
The suggestion from @cwendling isn`t bad, or not?

rbuj:
The suggestion is correct, since most [FPUs](https://en.wikipedia.org/wiki/Floating-point_unit) have a Round Unit, and you can choose the rounding mode after performing each FP operation.

raveit65:
> The suggestion is correct......

So, you like to change PR?

rbuj:
No, because we already use round function in several projects for a while, which is easier to understand at first glance and it's optimized for each arch.
```bash
mate-panel/panel-profile.c:     return (guint16) round (color.alpha * 65535);
mate-volume-control/gvc-channel-bar.c:        gdouble new_value = minimum + round (((value - minimum) / range) * 100) * (range / 100);
mate-volume-control/gvc-stream-applet-icon.c:        guint volume_percent = (guint) round (100.0 * volume / normal);
mate-volume-control/gvc-stream-status-icon.c:        guint volume_percent = (guint) round (100.0 * volume / normal);
src/gsd-media-keys-window.c:        height = round (height) - 1;
src/gsd-media-keys-window.c:        width = round (width) - 1;
src/gsd-media-keys-window.c:        x1 = round ((width - 1) * percentage);
src/gsd-media-keys-window.c:        icon_box_width = round (window_width * ICON_SCALE);
src/gsd-media-keys-window.c:        icon_box_height = round (window_height * ICON_SCALE);
src/gsd-media-keys-window.c:        volume_box_height = round (window_height * 0.05);
src/gsd-media-keys-window.c:        icon_box_x0 = round ((window_width - icon_box_width) / 2);
src/gsd-media-keys-window.c:        icon_box_y0 = round ((window_height - icon_box_height) / 2);
src/gsd-media-keys-window.c:        volume_box_x0 = round (icon_box_x0);
src/gsd-media-keys-window.c:        volume_box_y0 = round (window_height - icon_box_y0 / 2 - volume_box_height);
src/gsd-media-keys-window.c:        icon_box_width = round (window_width * ICON_SCALE);
src/gsd-media-keys-window.c:        icon_box_height = round (window_height * ICON_SCALE);
src/gsd-media-keys-window.c:        bright_box_width = round (icon_box_width);
src/gsd-media-keys-window.c:        bright_box_height = round (window_height * 0.05);
src/gsd-media-keys-window.c:        icon_box_x0 = round ((window_width - icon_box_width) / 2);
src/gsd-media-keys-window.c:        icon_box_y0 = round ((window_height - icon_box_height) / 2);
src/gsd-media-keys-window.c:        bright_box_x0 = round (icon_box_x0);
src/gsd-media-keys-window.c:        bright_box_y0 = round (window_height - icon_box_y0 / 2 - bright_box_height);
plugins/media-keys/msd-media-keys-window.c:        height = round (height) - 1;
plugins/media-keys/msd-media-keys-window.c:        width = round (width) - 1;
plugins/media-keys/msd-media-keys-window.c:        x1 = round ((width - 1) * percentage);
plugins/media-keys/msd-media-keys-window.c:        icon_box_width = round (window_width * ICON_SCALE);
plugins/media-keys/msd-media-keys-window.c:        icon_box_height = round (window_height * ICON_SCALE);
plugins/media-keys/msd-media-keys-window.c:        volume_box_height = round (window_height * 0.05);
plugins/media-keys/msd-media-keys-window.c:        icon_box_x0 = round ((window_width - icon_box_width) / 2);
plugins/media-keys/msd-media-keys-window.c:        icon_box_y0 = round ((window_height - icon_box_height) / 2);
plugins/media-keys/msd-media-keys-window.c:        volume_box_x0 = round (icon_box_x0);
plugins/media-keys/msd-media-keys-window.c:        volume_box_y0 = round (window_height - icon_box_y0 / 2 - volume_box_height);
plugins/media-keys/msd-media-keys-window.c:        icon_box_width = round (window_width * ICON_SCALE);
plugins/media-keys/msd-media-keys-window.c:        icon_box_height = round (window_height * ICON_SCALE);
plugins/media-keys/msd-media-keys-window.c:        label_box_width = round (window_width);
plugins/media-keys/msd-media-keys-window.c:        label_box_height = round (window_height * 0.175);
plugins/media-keys/msd-media-keys-window.c:        icon_box_x0 = round ((window_width - icon_box_width) / 2);
plugins/media-keys/msd-media-keys-window.c:        icon_box_y0 = round ((window_height - icon_box_height) / 2);
plugins/media-keys/msd-media-keys-window.c:        label_box_y0 = round (window_height - label_box_height / 2);
```

raveit65:
Yes and no in different postings, i have no idea what is right.
Needs to be reviewed by someone with more coding knowledge as me......

lukefromdc:
I just tested manually editing in the change to src/core/constraints.c and got a successful build with meson, which ran fine when installed. 

No idea if any checks need to be added to meson for this, but if not the best way to resolve the "silent merge conflict" might be to switch out the file for master's version and again apply these edits by hand. If that doesn't work, a new branch from master would be the only way I could do it, but I am not an expert in resolving git issues

lukefromdc:
I think I've figured out the purpose of this: if we were implicitly telling the compiler to round a float to an integer (this is still present in two more places according to comments) we are sending cleaner code to the compiler. What would make potentially lighter code would be if we could stay out of the FPU in the first place, but if that led to inferior results visually it would be a bad tradeoff for modern hardware. Metacity (from which marco was forked) dates back to the Pentium 4 era, with default settings it was a bit heavy for a Pentium 2 or 3 but worked.

cwendling:
> I think I've figured out the purpose of this: if we were implicitly telling the compiler to round a float to an integer (this is still present in two more places according to comments) we are sending cleaner code to the compiler.

No, @rbuj apparently meant to change the behavior: before this change, the value was clamped down (to the closest integer value towards 0, aka "truncating", removing the decimal part: -5.9 becomes -5, 5.9 becomes 5, 5.2 becomes 5), whereas with this change it rounds to the nearest integer (and away from 0, but that doesn't matter much: -5.9 becomes -6, 5.9 becomes 6, 5.2 becomes 5).

> What would make potentially lighter code would be if we could stay out of the FPU in the first place, but if that led to inferior results visually it would be a bad tradeoff for modern hardware. Metacity (from which marco was forked) dates back to the Pentium 4 era, with default settings it was a bit heavy for a Pentium 2 or 3 but worked.

I really don't think this should be a concern, it's *very* unlikely to have any kind of measurable performance impact, save maybe on 1960's hardware.

> No idea if any checks need to be added to meson build files for this

Well, this needs to link to libm for math functions. If it's not already there, it should be added or that won't work properly (yet it might work for some linkers as it's likely one of the dependencies already bring in libm, so some linkers will be happy, while others (depending also on configuration) will not unless there is an explicit dependency).

rbuj:
ping

rbuj:
It's only a `git rebase master`

rbuj:
> > I think I've figured out the purpose of this: if we were implicitly telling the compiler to round a float to an integer (this is still present in two more places according to comments) we are sending cleaner code to the compiler.
> 
> No, @rbuj apparently meant to change the behavior: before this change, the value was clamped down (to the closest integer value towards 0, aka "truncating", removing the decimal part: -5.9 becomes -5, 5.9 becomes 5, 5.2 becomes 5), whereas with this change it rounds to the nearest integer (and away from 0, but that doesn't matter much: -5.9 becomes -6, 5.9 becomes 6, 5.2 becomes 5).

We all know the difference between truncating and rounding. Do you know which one is better here?
 
> > What would make potentially lighter code would be if we could stay out of the FPU in the first place, but if that led to inferior results visually it would be a bad tradeoff for modern hardware. Metacity (from which marco was forked) dates back to the Pentium 4 era, with default settings it was a bit heavy for a Pentium 2 or 3 but worked.
> 
> I really don't think this should be a concern, it's _very_ unlikely to have any kind of measurable performance impact, save maybe on 1960's hardware.

I don't care, but... do you know how to prove it?
 
> > No idea if any checks need to be added to meson build files for this
> 
> Well, this needs to link to libm for math functions. If it's not already there, it should be added or that won't work properly (yet it might work for some linkers as it's likely one of the dependencies already bring in libm, so some linkers will be happy, while others (depending also on configuration) will not unless there is an explicit dependency).

is there any change?

cwendling:
> We all know the difference between truncating and rounding. Do you know which one is better here?

I don't.

> > > What would make potentially lighter code would be if we could stay out of the FPU in the first place, but if that led to inferior results visually it would be a bad tradeoff for modern hardware. Metacity (from which marco was forked) dates back to the Pentium 4 era, with default settings it was a bit heavy for a Pentium 2 or 3 but worked.
> > 
> > I really don't think this should be a concern, it's _very_ unlikely to have any kind of measurable performance impact, save maybe on 1960's hardware.
> 
> I don't care, but... do you know how to prove it?

It wouldn't really be a "proof", but I'd make a naive test code piece like that:
```c
#include <stdio.h>
#include <math.h>
#include <glib.h>

#define TIMES 5000000000lu
G_STATIC_ASSERT(TIMES < G_MAXULONG);

#define TIMEIT(rt, expr) \
  do { \
    volatile rt ret; \
    GTimer *timer = g_timer_new(); \
    for (unsigned long int i = 0; i < TIMES; i++) { \
      ret = (expr); \
    } \
    fprintf(stderr, "Calling `%s` %lu times took %gs (result is %d)\n", \
            G_STRINGIFY(expr), TIMES,  g_timer_elapsed(timer, NULL), ret); \
    g_timer_destroy(timer); \
  } while (0)

int main(int argc, char **argv)
{
  while (*++argv) {
    char *endptr;
    double value = strtod(*argv, &endptr);
    g_assert(! *endptr);
    int value_i10 = (int) (value * 10);

    TIMEIT(int, ((int) value)); /* original code (cast added) */
    TIMEIT(int, ((int) (value + 0.5))); /* manual round */
    TIMEIT(int, ((int) round(value))); /* PR's round() */
    TIMEIT(int, ((value_i10 + 5) / 10)); /* no-FPU rounding (?) */
  }

  return 0;
}
```

…and then build and run it in various ways:
```console
$ gcc -o /tmp/ms.out /tmp/ms.c -lm $(pkg-config --libs --cflags glib-2.0) -Wall -O0
$ /tmp/ms.out 1234.567
Calling `((int) value)` 5000000000 times took 12.2423s (result is 1234)
Calling `((int) (value + 0.5))` 5000000000 times took 12.6542s (result is 1235)
Calling `((int) round(value))` 5000000000 times took 21.2816s (result is 1235)
Calling `((value_i10 + 5) / 10)` 5000000000 times took 12.6601s (result is 1235)
$ gcc -o /tmp/ms.out /tmp/ms.c -lm $(pkg-config --libs --cflags glib-2.0) -Wall -O1
$ /tmp/ms.out 1234.567
Calling `((int) value)` 5000000000 times took 3.53481s (result is 1234)
Calling `((int) (value + 0.5))` 5000000000 times took 3.58011s (result is 1235)
Calling `((int) round(value))` 5000000000 times took 3.56202s (result is 1235)
Calling `((value_i10 + 5) / 10)` 5000000000 times took 3.52734s (result is 1235)
$ gcc -o /tmp/ms.out /tmp/ms.c -lm $(pkg-config --libs --cflags glib-2.0) -Wall -O2
$ /tmp/ms.out 1234.567
Calling `((int) value)` 5000000000 times took 3.57355s (result is 1234)
Calling `((int) (value + 0.5))` 5000000000 times took 3.52917s (result is 1235)
Calling `((int) round(value))` 5000000000 times took 3.51895s (result is 1235)
Calling `((value_i10 + 5) / 10)` 5000000000 times took 3.52162s (result is 1235)
$ gcc -o /tmp/ms.out /tmp/ms.c -lm $(pkg-config --libs --cflags glib-2.0) -Wall -O3
$ /tmp/ms.out 1234.567
Calling `((int) value)` 5000000000 times took 3.54635s (result is 1234)
Calling `((int) (value + 0.5))` 5000000000 times took 3.53554s (result is 1235)
Calling `((int) round(value))` 5000000000 times took 3.51218s (result is 1235)
Calling `((value_i10 + 5) / 10)` 5000000000 times took 3.55056s (result is 1235)
```

What we can see is that:
* at `-O0`, calling `round()` is "slow".
* at `-O1` and above, all results are basically identical.
* working on integers show not difference.

Anyway, even without measurements this code path here really don't seem hot to me, so I hardly think any of this actually matters.  I'd just use whatever code is best readable and maintainable without worrying about performance.  My comment about using `round()` is merely due to it not really be needed, and that there was a problem linking to libm at some point.  But either way is fine, and which one to use don't matter.

The only question is whether or not rounding is the thing to do, or if floor()ing was better.

rbuj:
why did you take a run test? do you know how to profile? 😄

rbuj:
CPU profiling example: https://github.com/mate-desktop/mate-settings-daemon/pull/311

cwendling:
I'm not gonna profile Marco, specially not over this :).  This code is for computing the aspect ratio constraint for resizing windows, and it's doing orders of magnitude more complex computations all over the place; that line won't ever show in a profile.  And profiling non-calls is a lot less fun, so we'd have to make sure it's a real function that gets called, which by itself is gonna make it slower.

But if you wanna do that, please, do :)

PR #480, opened 2019-04-08
Fix cast from non-struct type to struct type

sc0w:
Fixes Clang static analyzer warnings:

```
core/xprops.c:761:9: warning: Casting a non-structure type to a structure type and accessing a field can lead to memory access errors or data corruption
  raw = (xPropWMHints*) results->prop;
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~

core/xprops.c:883:9: warning: Casting a non-structure type to a structure type and accessing a field can lead to memory access errors or data corruption
  raw = (xPropSizeHints*) results->prop;


core/async-getprop.c:277:11: warning: Casting a non-structure type to a structure type and accessing a field can lead to memory access errors or data corruption
  reply = (xGetPropertyReply *)
          ^~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/core/async-getprop.c
+++ b/src/core/async-getprop.c
@@ -274,7 +274,7 @@ async_get_property_handler (Display *dpy,
 #endif
 
   /* (kind of a silly as we know sizeof(xGetPropertyReply) == sizeof(xReply)) */
-  reply = (xGetPropertyReply *)
+  reply = (xGetPropertyReply *) (void *)
     _XGetAsyncReply (dpy, (char *)&replbuf, rep, buf, len,
                      (SIZEOF (xGetPropertyReply) - bytes_read) >> 2, /* in 32-bit words */
                      False); /* False means expecting more data to follow,
--- a/src/core/xprops.c
+++ b/src/core/xprops.c
@@ -758,7 +758,7 @@ wm_hints_from_results (GetPropertyResults *results,
 
   hints = ag_Xmalloc0 (sizeof (XWMHints));
 
-  raw = (xPropWMHints*) results->prop;
+  raw = (xPropWMHints*) (gpointer) results->prop;
 
   hints->flags = raw->flags;
   hints->input = (raw->input ? True : False);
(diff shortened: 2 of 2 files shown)


muktupavels:
Hmm, from commit message it is not clear (to me) if this really fixes something or just avoids this warning.

> Casting a non-structure type to a structure type and accessing a field can lead to memory access errors or data corruption

Might be wrong, but casting first to `void *` does not invalidate warning, does it? It just looks like trick to hide warning.

lukefromdc:
Should we hold this? @muktupavels says this doesn't FIX anything at all but only hides a warning.

sc0w:
Fixes the warnings and there is no issues

For me this PR is ready to merge

vkareh:
Looking at this is a bit concerning, this hides a warning but will it cause some unknown issue down the line? what exactly does this change do in terms of code?

cwendling:
> ktupavels says this doesn't FIX anything at all but only hides a warning.

It does indeed.  But looking a little bit at it, I'd say the code is right in its casting, and that it's simply unfortunate that e.g. `_XGetAsyncReply()` returns `char*` rather than `void*` (maybe it's some super old pre-ANSI C, or just some pre-ANSI practices…) which trigger this warning.

I don't have an opinion on whether these warnings should be effectively silenced, but I'd say the changes are OK.

>  what exactly does this change do in terms of code behavior?

Nothing.  But I believe the code was correct anyway.

lukefromdc:
What do we want to do with this one? Merge it or close it?

cwendling:
> Might be wrong, but casting first to `void *` does not invalidate warning, does it? It just looks like trick to hide warning.

It is meant to hide the warning, and it should likely do so.  The thing is that those functions return `char*` but *seem* to mean `void*`, and thus a ANSI compiler can legitimately warn about the conversion.  Well, here the conversion being already explicit, it's more questionable, and thus it's questionable whether an additional cast to `void*` would always silence the warning or whether a future version of the tools would get more "clever" and warn again.

IMO this can be merged so the warning is not bothering people again, but I'm not confident this is a "fix", in that the compiler could just as well warn with this as well trying to be clever.

vkareh:
Reading https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-function-type it suggests using a similar technique to supress the warnings. I think we should merge this. Even GLib does the same: https://gitlab.gnome.org/GNOME/glib/commit/039fa6897bb4c5cd30caec8ee3add1a4ef612f76

muktupavels:
How `-Wcast-function-type` is/was related to this?

Issue #479, opened 2019-04-06
 Alt+tab: icons are too big


osch:
After #447 the alt+tab icons are too large on my lowdpi screen (now they are: 1.6cm, 52 pixels).

The icon size seems to be hardcoded in `src/include/common.h` to 48. If I change this value to 32 (see diff below) the icons have the smaller size as they used to be before #447 (1.1cm, 36 pixels).

```
diff --git a/src/include/common.h b/src/include/common.h
index 8193730..83b750a 100644
--- a/src/include/common.h
+++ b/src/include/common.h
@@ -304,2 +304,2 @@ struct _MetaButtonLayout
-#define META_ICON_WIDTH 48
-#define META_ICON_HEIGHT 48
+#define META_ICON_WIDTH 32
+#define META_ICON_HEIGHT 32
```

What do you think about making this hardcoded value configurable? I could try to prepare a pull request for this, what do you think?

MATE Version: 1.22.0
See also: #333 

Screenshots:
![alttab00](https://user-images.githubusercontent.com/42665/55669521-94f64c80-5878-11e9-8427-f89a1ae5505b.png)
![alttab01](https://user-images.githubusercontent.com/42665/55669522-9889d380-5878-11e9-9116-e2a6dbf7fea4.png)

PR #482, opened 2019-04-12
make icon size configurable

osch:
see #479

### DIFF
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -812,3 +812,25 @@ meta_invalidate_default_icons (void)
   g_slist_free (windows);
 }
 
+void
+meta_invalidate_all_icons (void)
+{
+  MetaDisplay *display = meta_get_display ();
+  GSList *windows;
+  GSList *l;
+
+  if (display == NULL)
+    return; /* We can validly be called before the display is opened. */
+
+  windows = meta_display_list_windows (display);
+  for (l = windows; l != NULL; l = l->next)
+    {
+      MetaWindow *window = (MetaWindow*)l->data;
+
+      meta_icon_cache_invalidate (&(window->icon_cache));
+      meta_window_update_icon_now (window);
+    }
+
+  g_slist_free (windows);
+}
+
--- a/src/core/iconcache.c
+++ b/src/core/iconcache.c
@@ -523,6 +523,10 @@ meta_icon_cache_init (MetaIconCache *icon_cache)
   icon_cache->wm_hints_dirty = TRUE;
   icon_cache->kwm_win_icon_dirty = TRUE;
   icon_cache->net_wm_icon_dirty = TRUE;
+
+  icon_cache->wm_hints_dirty_forced = FALSE;
+  icon_cache->kwm_win_icon_dirty_forced = FALSE;
+  icon_cache->fallback_icon_dirty_forced = FALSE;
 }
 
 static void
@@ -553,6 +557,18 @@ void
 meta_icon_cache_free (MetaIconCache *icon_cache)
 {
   clear_icon_cache (icon_cache, FALSE);
+}
+
+void
+meta_icon_cache_invalidate (MetaIconCache *icon_cache)
+{
+  icon_cache->wm_hints_dirty = TRUE;
+  icon_cache->kwm_win_icon_dirty = TRUE;
+  icon_cache->net_wm_icon_dirty = TRUE;
+
+  icon_cache->wm_hints_dirty_forced = TRUE;
+  icon_cache->kwm_win_icon_dirty_forced = TRUE;
+  icon_cache->fallback_icon_dirty_forced = TRUE;
 }
 
 void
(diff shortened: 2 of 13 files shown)


raveit65:
This works very nice and icons are scaling fine with a HIDPI monitor and a very huge icon size. Screenshot is with 420px icon-size :)
Bildschirmfoto zu 2019-04-14 14-30-06
Changing icon-size is also possible with using thumbnail-previews, which is great.
But i think in case of using thumbnails we need a size limit, because icons don't display if previews are smaller than the chosen icon size.
Than we get and errors in .xsession-errors log like this, in my case when icon-size is > 220px for windows with default size.
```
(marco:17563): GdkPixbuf-CRITICAL **: 14:12:12.860: gdk_pixbuf_composite: assertion 'dest_y >= 0 && dest_y + dest_height <= dest->height' failed
```
Beside from that this is a very nice feature.

sc0w:
yes, I think it needs limits to big and small icons, it prevents unexpectedly behavior if user put a lot of bigger/smaller icon by mistake

and I think @alexarnaud will like this new feature for accessibility reasons :)

osch:
The error `gdk_pixbuf_composite: assertion` is not new: you can also have this error with current marco if you resize the window such that e.g. the height is much smaller than the width: see the following screenshot from alt+tab popup: both window thumbnails on the left are firefox windows but the first one does not contain an icon because the icon does not fit into the provided height and therefore causes the assertion.

image

raveit65:
Cool,
last commit does what it promised and scales down the icon, if the thumbnail is smaller than icon-size.
But 1024px seems to be very large. With 1024px the icon doesn't fit in my UHD-monitor (3840x2160px).
Bildschirmfoto zu 2019-04-14 22-12-14
For which videowall setup you need such a big icon?
I think a limit of 480px should be sufficient. 640x480 is smallest monitor size.

sc0w:
I think the max can be calculated depending of the screen monitor dimensions

osch:
I personally only need icon size 32 so I'm also fine with max size 480. The motivation for choosing 1024 as max value was not "which max value would be enough", it was instead the question "which max value is beyond resonable max value and does not harm the principle functionality of the appliation". And if you say 480 is reasonable max value, than I'm glad that 1024 is not below. IMHO it's always best to let the user decide. Who knows how big screens will be in the future? But as I said: I'm also fine with max value 480. 

I also think it could be possible to calculate max value from screen dimensions, but at least a fixed max value must be provided in the gschema.xml . But where's the problem? If the user chooses a max value that does not make sense, the user will recognise this and change the value back.

rbuj:
META_MIN_ICON_SIZE should be 8.
Common png icon sizes: 8, 16, 22 , 24, 32, 48. (svg: multiple of 16=16..64)
png logo size is multiple of 128 (usually 256). (svg: multiple of 128=128...512)

rbuj:
In caja:
64 is the default preview size (it can be set with dconf).
Icon scale: 48* is the base (1:1). 16* is 33%, 24* is 50%, 32* is 66%, 72 is 150%, 96 is 200%, 192 is 400%.

osch:
Min icon size 6 works perfectly here on my low dpi screen. I don't see any problems with min icon size 6. What's wrong if a user decides to use icon size 6 and it does work? 

I cannot reproduce problems with icon size 128 or larger here. See the following screenshot with icon size 128:

image

Perhaps I'm missing something and this is not the "window switcher preview"?

rbuj:
@sc0w 
It's a great improvement. It requires fixing some icons in the **High Contrast Inverse** theme, since some icons appears blurred. Fix the frame color of the window selector. Create the symbolic link: mate-desktop -> mate.

Captura de pantalla a 2019-04-15 11-48-35


off-topic:
**High Contrast Inverse** theme it's not detected as [high contrast theme in firefox](https://support.mozilla.org/en-US/kb/accessibility-features-firefox-make-firefox-and-we#w_using-a-high-contrast-theme)

lukefromdc:
One possiblity is that my icon theme(a custom one) by not providing some of the larger icons may have been an issue. Since I normally run compiz and my own GTK and icon themes, I might not notice some things in marco and some icon theme or GTK theme dependent things do act differently.

raveit65:
@rbuj 
min size =6px is OK if using 
```
[rave@mother Schreibtisch]$ gsettings get org.mate.Marco.general compositing-fast-alt-tab
true
```
, which was the first intention of pull request.
Bildschirmfoto zu 2019-04-17 14-06-07
Well, the icon is small but you can see which icon is it.

It seems there is a different with the icon-size if using thumbnails.
I can confirm that 6 or 8px is nearly a point, or an icon without any meaning.
For me everything < 16px doesn't make any sense when using thumbnails.

raveit65:
@rbuj 
Is the min/max icon-size really a show stopper for you?
Or can we merge it like it is?
6px is working for me if using non thumbnails.

rbuj:
The desktop [UI standard is still 96 dpi](https://www.eizo.se/knowledge/monitor-expertise/understanding-pixel-density/).
```
pixels = 6;
dpi = 96;
mm = 25.4 * pixels / dpi;
mm = 1.5875
```
Recomended [viewing distance](http://ergo.human.cornell.edu/ergoguide.html): 1 arm (0.8m)

I can't distinguish an image of 1.6x1.6mm at 0.8m.

raveit65:
And using 8px makes you mathematics results better?
My eyes are not the best but i can recognize the icons if i don't use using thumbnails.
Please don't forget the merge this PR, i have to delete it from my TODO list

sc0w:
I think it is a good idea to do the same in the workspace switcher :)

PR #487, opened 2019-04-23
window: make docks and spashscreens appear focused

raveit65:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1701799
MATE Menu Bar often has unreadable text with none MATE themes, which support backdrop state.

Set the NET_WM_STATE_FOCUSED property on windows of type dock or
spashscreen so that they don't get the state GTK_STATE_FLAG_BACKDROP
set by default.

Based on:
https://gitlab.gnome.org/GNOME/metacity/commit/b3ef887
origin xfwm4 commit:
https://git.xfce.org/xfce/xfwm4/commit/?id=0feb29e78bb3

For testing use adwaita theme which has backdrop state support.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -8661,7 +8661,15 @@ meta_window_appears_focused (MetaWindow *window)
       meta_window_foreach_transient (window, transient_has_focus, &focus);
       return focus;
     }
-  return window->has_focus;
+
+  if (window->has_focus)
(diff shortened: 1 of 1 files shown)


raveit65:
Sadly, this doesn't work.
Need to investigate why...

raveit65:
Now it works. Please Test.

Issue #492, opened 2019-05-07
Make alt+tab window layout customizable/wider


stefanct:
Since we got configurable icon sizes with #482 maybe someone familiar with the code already wants to fulfill a wish of mine ;) This is also somewhat related to  #68 depending on how we want to resolve this issue.

#### Expected behaviour
An alt+tab selection window that is large enough to properly display window titles if there is enough (horizontal) space to make the selection window bigger.

#### Actual behaviour
I think there is a hard limit of numbers of icons per row in the alt+tab window (5 in my version). Even with properly sized icons this is unnecessarily small and on high resolutions the alt+tab window is ridiculously small in comparison to the overall screen. This would be not an issue in itself (apart from aesthetics) since it uses a multi-row layout for the icons. However, it reduces the functionality because the title of the currently selected application/window could have way more space and is abbreviated if they are too long.

#### Steps to reproduce the behaviour

Open a lot of windows with long title texts and press alt+tab

#### MATE general version

1.16.2

#### Package version

1.16.1

#### Linux Distribution

Debian stable/stretch

#### Link to downstream report of your Distribution

Can do but I think think this should rather be discussed upstream since it is a design decision not a bug...

vkareh:
Fixed by #497 

Since this change you can configure the number of columns, the icon size, and whether the popup expands to fit the title.

PR #497, opened 2019-05-14
Make alt+tab max columns configurable and consider window label width

osch:
New config parameter `org.mate.Marco.general.alt-tab-max-columns` to configure *Maximum number of columns in alt-tab popup window*. The popup window may be expanded up to the full width to fit window labels if less columns are displayed.
See also #492


sc0w:
Thanks, all the warnings I mentioned seem to be fixed

osch:
My PR consists of two parts:

1.) max columns in popup window can be configured
2.) The popup window may be expanded up to the full width to fit window labels if less columns are displayed.

So it could be that the extra space is intentional because of 2.) To check this: Can you proceed with alt+tab in your screenshot to see if there are longer window titles?

src/ui/tabpopup.c
```diff
@@ -378,28 +379,59 @@ meta_ui_tab_popup_new (const MetaTabEntry *entries,
...
   gtk_label_set_ellipsize (GTK_LABEL (popup->label), PANGO_ELLIPSIZE_END);
 
-  /* Limit the window size to no bigger than screen_width/4 */
-  if (max_label_width>(screen_width/4))
+  int default_window_width = 0; /* 0 == small as possible, truncate label */
+
+  if (consider_label_width && top <= 1 && left < width)
     {
-      max_label_width = screen_width/4;
-    }
```
vkareh: I wonder if we can somehow limit the max width to a fraction of the screen, say, 80% or 90% of it, and automatically wrap - so basically if the widget is too large for the user's screen, instead of honoring their setting, we add a new row instead.

vkareh:
> Can you proceed with alt+tab in your screenshot to see if there are longer window titles?

Yes! That was it, thanks!

Screenshot at 2019-05-24 06-52-40

osch:
> However, I have to say that that the label width thing is a bit jarring.

I could implement another configuration option to turn off the "label width thing". I personally very much like this feature. By the way: in the old code there already was code that tried to make the popup larger to some extend (and adding space) to fit the label width, however this code was broken (perhaps duet to gtk2/3 changes/updates) and always calculated a label width of 0 and therefore always truncated the label to the column size, which make the label nealry unusable if there are only 2-4 icons and small icon size (32).

src/ui/tabpopup.c
osch: This could also be possible, but I very much like the feature to configure the max number columns myself. So a solution could be to allow max-columns = 0 and this could mean: 80% of screen size.

osch:
In the last commit I implemented an additional config parameter `org.mate.Marco.general.alt-tab-max-empty-columns` to configure the number of empty columns that are added up to `alt-tab-max-columns` to fit long window labels.

The default values are: `alt-tab-max-columns=5` and `alt-tab-max-empty-columns=0`. With these values the behaviour is exactly as before. 

All the following examples have `alt-tab-max-columns=5` and `icon-size=48`.

Example with `alt-tab-max-empty-columns=0`:
image

Example with `alt-tab-max-empty-columns=1`:
image

Example with `alt-tab-max-empty-columns=2`:
image

Example with `alt-tab-max-empty-columns=100`:
image

Example with `alt-tab-max-empty-columns=100`:
image

src/org.mate.marco.gschema.xml
```diff
@@ -202,6 +202,18 @@
         <summary>Icon size</summary>
         <description>Size of the application icons displayed in alt-tab popup window. The screen's scale factor is applied to this value.</description>
     </key>
+    <key name="alt-tab-max-columns" type="i">
+        <range min="1" max="1024"/>
+        <default>5</default>
+        <summary>Maximum number of columns in alt-tab popup window</summary>
+        <description>The popup window may be expanded by empty columns up to the maximum columns width to fit window labels if less columns are displayed and if config option 'alt-tab-max-empty-columns' has a value > 0.</description>
+    </key>
+    <key name="alt-tab-max-empty-columns" type="i">
```
vkareh: Would it make sense to change this to `alt-tab-expand-to-fit-title` or something like that and make it a `boolean`? The concept of adding arbitrary columns for padding seems a bit kludgy. In `false` mode, it would be the current behavior (expand only to fit icons), in `true` mode expand the alt-tab window to fit window titles.

src/org.mate.marco.gschema.xml
```diff
@@ -202,6 +202,18 @@
         <summary>Icon size</summary>
         <description>Size of the application icons displayed in alt-tab popup window. The screen's scale factor is applied to this value.</description>
     </key>
+    <key name="alt-tab-max-columns" type="i">
+        <range min="1" max="1024"/>
+        <default>5</default>
+        <summary>Maximum number of columns in alt-tab popup window</summary>
+        <description>The popup window may be expanded by empty columns up to the maximum columns width to fit window labels if less columns are displayed and if config option 'alt-tab-max-empty-columns' has a value > 0.</description>
```
vkareh: This description is a bit confusing... maybe change to something like:

> The popup window will be expanded to fit up to these many entries per row.

No need to mention the window titles, as those would be handled separately by the next variable.

osch: Ok, I changed it as you suggested, see next commit 06129b4 below.

src/org.mate.marco.gschema.xml
osch: With a boolean value we will loose the possibility to add _some_ space for long labels but not up to the maximum according to max-columns. Instead of columns one could specify extra space in abstract pixels or percentage relative to max popup width of how much space could be added for long labels.

src/ui/tabpopup.c
osch: IMHO in general it should be possible to let the user limit the size of the popup and let not the screen determine this limit: The label text in the popup is centered. The user could have problems to read the label while switching if the screen width is very large and therefore the distance from the left border of the popup to the beginning of the label text could become also very large.

src/org.mate.marco.gschema.xml
vkareh: hmm... not sure I fully understand. The Alt+Tab window should expand to either fit the icons or to fit the longest label, not sure why any intermediate step is necessary. Adding arbitrary space/pixels/padding seems like a theme issue, not a user setting.

osch: It's not about extra arbitrary space, it's about how much space is added for longer labels. The amount of added space for longer labels is not only limited by the label length, it's also limited by the max-columns parameter (and perhaps also by max-empty-columns).

Perhaps it may be that the user wants to limit the amount of empty space added around icons if larger labels are present. You yourself were first irritated by the large amount of space added, this space could be limited by the max-empty-column parameter. Did you see my screenshots above demonstrating how extra space my be limited using the max-empty-column parameter? If we have only a single flag this flexibility will be lost. Anyway I can also live with a flag.

osch: Ok, I changed it as you suggested, see commit  9f66e47 below. We will loose some flexibility in configuring alt-tab layout but perhaps it is now easier to understand.

osch:
After the last commit 9f66e47 this PR contains the following:

1.  New config parameter `org.mate.Marco.general.alt-tab-max-columns` to configure maximum number of columns in alt-tab popup window. 

2. New config parameter `org.mate.Marco.general.alt-tab-expand-to-fit-title` to expand the alt-tab popup window to fit longer window titles. The popup window may be expanded up to the width determined by the config parameter `alt-tab-max-columns`.

Default values are `alt-tab-max-columns = 5` and `alt-tab-expand-to-fit-title = false`, this behaves exactly as it was before this PR.

vkareh:
Thanks so much for seeing this through! I'm quite eager to test this (will do tomorrow morning when I'm at the office).

raveit65:
Any chance to change that?
I think [PATCH 10/10] can be squashed into [PATCH 06/10], or not?
And [PATCH 08/10] and [PATCH 07/10] can be squashed into [PATCH 06/10], or not?
Those are only corrections from previous commit.
Having less commits is better for bisceting git history.

vkareh:
@raveit65, can we just squash the entire thing? It's a single feature anyway. From a history point of view this would be far easier as a single commit...

raveit65:
@vkareh 
I am fine with one single commit.
I can do that during merging after i have tested the PR........today :)

raveit65:
Merged as one single commit.
https://github.com/mate-desktop/marco/commit/604fa96f1026fbffd4dcf21a864076d068e6f56a

PR #499, opened 2019-05-16
Handle mouse clicks in Alt+Tab Popup

### LINKED ISSUE #364: Alt-tab Application switcher should allow clicks [$50]
the alt-tab application switcher does not allow clicks

https://github.com/mate-desktop/marco/pull/330 fixes it so that you can use keys so that you do not have to cycle things linearly. but it should also support clicks. windows, mac and gnome all support this so we should add it to mate.

#### Expected behaviour
If you have lots of apps, application switcher should allow you to click the applications to show them

#### Actual behaviour
You must use tab or shift tab to cycle linearly which is not useful when you have many apps open 

Hit alt tab with many windows open. notice clicking a window in the switcher just closes the switcher.

#### MATE general version
tested on ubuntumate 16.x and 17.x both show this missing feature

#### Package version
n/a

#### Linux Distribution
Ubuntu 16.x and 17.x

https://bugs.launchpad.net/ubuntu-mate/+bug/1739242

osch:
Now it's possible to move the focus rectangle in the alt+tab popup by mouse clicks.
Fixes #364.

### DIFF
--- a/src/core/display-private.h
+++ b/src/core/display-private.h
@@ -128,6 +128,8 @@ struct _MetaDisplay {
 
 	guint static_gravity_works: 1;
 
+        guint tab_popup_mouse_pressed : 1;
+
 	/*< private-ish >*/
 	guint error_trap_synced_at_last_pop: 1;
 	MetaEventQueue* events;
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -388,6 +388,8 @@ meta_display_open (void)
   /* FIXME copy the checks from GDK probably */
   the_display->static_gravity_works = g_getenv ("MARCO_USE_STATIC_GRAVITY") != NULL;
 
+  the_display->tab_popup_mouse_pressed = FALSE;
+
   meta_bell_init (the_display);
 
   meta_display_init_keys (the_display);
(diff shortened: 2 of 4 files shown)


osch:
this is strange... perhaps it has something todo with screen scale factor, because here X11 coordinate and gtk coordinate come together... @vkareh: what screen scale factor do you have? I am going to investigate this further.

vkareh:
I have it set to 2, that's a good point, it might need correction for scaling.

But I wonder if the window icons cab be wrapped in a widget type that accepts click signals and pass the window id to the callback, that way you won't need to calculate the position...

osch:
> But I wonder if the window icons cab be wrapped in a widget type that accepts click signals and pass the window id to the callback, that way you won't need to calculate the position...

This was also my first idea, however I tried this approach first and it didn't work, because the events are not propagated to the gtk widgets. See also #437 where this approach (by using `gtk_event_box`) was also tried. There seems to be complicated event filtering code involved which I didn't dare to change. Then I detected that also the keyboard events are managed directly from x11 and not propagated to the gtk widgets (see invocation of `meta_display_process_key_event` in `marco/src/core/display.c`). Therefore I implemented the mouse clicks similiar and the current solution of this PR seems to be rather easy now. As I already wrote: I wil investigate it further to fix the screen scaling issues...

lukefromdc:
@vkareh , did this fix the problems you reported?

vkareh:
@osch, you should claim the bounty: https://www.bountysource.com/issues/52896480-alt-tab-application-switcher-should-allow-clicks

phocean:
It's strange that it's not working in 19.10, isn't it?

lukefromdc:
No it is not: Ubuntu 19.10 has mate 1.22, the fix for this was merged to master (which is 1.23) and I don't see any cherrypicks of this to version 1.22. If this can be cherrypicked to 1.22, only then can Ubuntu and other distros that don't ship MATE development versions pick it up.

phocean:
I see, thanks. That's sad to miss this nice feature.

raveit65:
We only merge bugfixes to  stable branch.
This is definitely a new feature which is for next Mate release
But you can ask your ubuntu maintainer to use it as patch for 19.10.

PR #503, opened 2019-05-23
Backport invisible resize borders

vkareh:
So far I'm happy with the result of this. It's a sequential backport of Metacity commits to get invisible resize borders working. I've made sure to preserve Marco's own HiDPI logic this time.

Fixes mate-desktop/mate-themes/issues/249
Fixes #481
Fixes #248

PR #504, opened 2019-05-24
Backport/1 reset visible borders

vkareh:
1. #503

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -83,6 +83,7 @@ marco_SOURCES = \
 	core/workspace.h \
 	core/xprops.c \
 	include/xprops.h \
+	core/common.c \
 	include/common.h \
 	include/core.h \
 	include/ui.h \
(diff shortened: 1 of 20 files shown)


raveit65:
@yetist 
All this PRs failed to build because of generating ccpcheck results in fedora build.
I think it is this line which causes the issue.
https://github.com/mate-desktop/mate-dev-scripts/blob/master/travis/gen-index.sh#L136
But why?
Travis building of first huge PR was fine https://github.com/mate-desktop/marco/pull/503
See https://travis-ci.org/mate-desktop/marco/builds

sc0w:
> @yetist
> All this PRs failed to build because of generating ccpcheck results in fedora build.
> I think it is this line which causes the issue.

I think it happens probably because the same commit is in two PRs, I think travis will be ok in master when merged.

raveit65:
> I think it happens probably because the same commit is in two PRs, I think travis will be ok in master when merged.

This shouldn't happen. It's normal business to split PRs in smaller ones.
If ccpcheck implementation can't handle this that this should be removed from fedora builds at travis-CI.

sc0w:
I noticed if the job is restarted here, in PR https://github.com/mate-desktop/marco/pull/503 the job of the same commit is restarted too, I think this is a travis issue

@yetist thoughts ?

PR #513, opened 2019-05-25
travis: use Debian Buster which has 1.20

sc0w:


raveit65:
thanks you for email spam!

PR #517, opened 2019-05-31
Fix gradients

### LINKED ISSUE #436: HiDPI anomalies 1.20
I just got a "new" laptop Dell XPS 15 9550 (2015 version) with a 4K display. I am running Debian 9/Mate 1.20.

If I use any of the Marco compositors, things look pretty good except for an annoying horizontal line through each window titlebar (see picture).

[Pic 1](http://realityisreality.com/mate-marco-compton.png)

If I use Compiz, there is no line through the window titlebar, but the text in the title bar is not scaled (i.e. it is tiny) (see picture).

[Pic 2](http://realityisreality.com/mate-compiz.png)

Any advice/workaround much appreciated!

### LINKED ISSUE #496: metacity-decoration edge glitches and other issues with HIDPI
#### Expected behaviour
clear and defined edges with all metacity themes if using HIDPI resolutions.
Correct color gradient with Traditional themes

#### Actual behaviour
Edges are undefined with double border lines
 Color gradient is broken with Traditional themes

Using Traditional, Submarine, GreenLaguna and Menta themes with marco VM with a HIDPI resolution.

#### MATE general version
1.21.1

#### Package version
1.21.x, 1.20.x and master

#### Linux Distribution
fedora 30

no need to file out a report against myself, i am informed :)

First, all this issues doesn't exists with normal monitor resolutions.
Metacity WM display all this themes well with HIDPI.

TraditionalOK with marco:
traditionalOK-marco
TraditionalOK with metacity:
traditionalOK-metacity
GreenLaguna with marco:
greenlaguna-marco
GreenLaguna with metacity:
greenlaguna-metacity
Blue-Submarine with marco:
blue-submarine-marco

Menta themes have edge glitches too, but not always. Mostly it happaens if using tiling modes.
I will make screenshots when it happens.

vkareh:
Fixes #436
Fixes #496

### DIFF
--- a/src/ui/gradient.c
+++ b/src/ui/gradient.c
@@ -22,40 +22,17 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301, USA.  */
 
-#include "gradient.h"
-#include "util.h"
+/**
+ * SECTION: gradient
+ * @title: Gradients
+ * @short_description: Metacity gradient rendering
+ */
+
+#include "config.h"
+
 #include <string.h>
 
-#include <gdk/gdk.h>
-
-/* This is all Alfredo's and Dan's usual very nice WindowMaker code,
- * slightly GTK-ized
- */
-static GdkPixbuf* meta_gradient_create_horizontal       (int             width,
-                                                         int             height,
-                                                         const GdkRGBA  *from,
-                                                         const GdkRGBA  *to);
-static GdkPixbuf* meta_gradient_create_vertical         (int             width,
-                                                         int             height,
-                                                         const GdkRGBA  *from,
-                                                         const GdkRGBA  *to);
-static GdkPixbuf* meta_gradient_create_diagonal         (int             width,
-                                                         int             height,
-                                                         const GdkRGBA  *from,
-                                                         const GdkRGBA  *to);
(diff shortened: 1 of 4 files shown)


lukefromdc:
In the second commit we have

`GSList *tmp;`

In the `while (tmp != NULL)` block we have

`tmp = tmp->next;`

Does this reliably clear *tmp or do we have a memory leak here? The cairo pattern is destroyed when we are done with it so no problem there.

vkareh:
Thanks for testing. For me, I see a thin line between the titlebar and the menu bar on *Menta and Ambiant* when on HiDPI, along with other random glitches. This PR fixes those as well.

vkareh:
That list is a reference to `spec->color_specs`, which is not owned by that function. I don't think it should be freed.

vkareh:
With regards to `tmp`, it can only exit the `while` once it's `NULL`, so yes, that variable gets cleared

lukefromdc:
Thanks, just wanted to make sure about the memory management. I'm not an expert on that but I do try to keep an eye out for introducing new problems there.

raveit65:
I will re-open https://github.com/mate-desktop/marco/issues/496 because glitches with borders when windows are in tiling mode aren't fixed. Eg. our default theme Menta.

PR #518, opened 2019-06-05
cherry-pick invisible borders to 1.22

vkareh:
I had to add 39191b621602839318e21d2727a1489a8c6ddde7, 7666add4e9b87504655546dda320d640e87ad938, and a53ba395d9baf432be480a89b7b14d64d3fb1069 from `master` to avoid conflicts


raveit65:
Honestly, i like to have a fix for gtk-decorator of compiz-reloaded first.
I can't use a 1.22.x tarball in f30/29 without breaking existing compiz-reloaded version.

vkareh:
Makes perfect sense! let's keep this PR open until we have a viable solution for compiz-reloaded

lukefromdc:
This will need a new release version, and note that we will be breaking API on a micro version number. Distros will need to know to rebuild compiz-reloaded to accomodate this

vkareh:
I'm wondering that we should not cherry-pick this for 1.22 and leave it for 1.24 only...
Maybe we can create a separate version tag, like 1.22.99, with all these commits squashed so that distros can either use that or patch a single commit... Any better ideas?

sc0w:
I don't agree to do a lot of changes in code over a stable release, with probably more hidden issues

lukefromdc:
If we cherrypick this, we break abi on a micro version, and force compiz-reloaded
also to cherrypick the changes there to 0.8.16 and release in lockstep. Certainly
new issues will crop up, and one or more distros will push the new marco release
but not the new compiz release. It's one thing for this to happen on Debian Stable,
quite another to have this happen in say, Debian Stable.

I agree with a 1.22.90 series (like hoe GNOME prereleases) that distros can use or
not and know is a major change.

raveit65:
> I don't agree to do a lot of changes in code over a stable release, with probably more hidden issues

Why?
And why are you always so negative and say something without giving a good reason.
Feel free to review other people PR's if you see issues....

For me this is a bugfix an people are waiting a long lime for it.

raveit65:
Guys, as OWNER of compiz-reloaded i see no problems to merge this to 1.22 branch, because compiz-reloaded is only used by official fedora and OpenSuse, and both distro maintainers are informed and happy to rebuild compiz.
So this isn't really a problem.

lukefromdc:
OK, that means all known distros using compiz-reloaded know of this
and presumably are OK with it. Anyone else then presumably is building
from source. Ideally compiz-reloaded would branch 0.8.18 or also backport
(should be a direct cherrypick)to 0.8.16 to ensure builders get the support
for this. Then it should be no problem, and itns IS a bugfix. A nasty bug and
a complex fix, so no perfext soluttions.

raveit65:
> ........Ideally compiz-reloaded would branch 0.8.18....

I will talk with soreau about a 0.8.16.2 release. But i noticed latest compiz-reloaded master has already an API change which needs an update/rebuild of the plugins packages.
So, we have another variable in this game which makes an simple new release of only compiz-core impossible.

lukefromdc:
In compiz-reloaded this affects only the gtk-window-decorator and does not change anything in compiz-core.h . Therefore the changes to support this should be able to be cherrypicked to compiz 0.8.16 without touching the core changes that break ABI for plugins.

raveit65:
PR for compiz-reloaded is merged. https://gitlab.com/compiz/compiz-core/merge_requests/139

raveit65:
Only a rebase from 1.22 to get version bump commit for better testing.

lukefromdc:
One point on use of compiz-reloaded: Debian now also uses it

flexiondotorg:
@lukefromdc When you say Debian now uses it, what do you mean exactly? Is there a desktop environment including it in their meta packages?

raveit65:
@lukefromdc 
And which debian version?
Is it an official debian package?
And keep in mind that next stable debian release ignore our current stable release and will use 1.20.
So they will never use a 1.22 package.

sc0w:
@raveit65 

debian have compiz reloaded 0.8.16.1 in buster (testing freezed / future stable) and sid

> And keep in mind that next stable debian release ignore our current stable release and will use 1.20.
> So they will never use a 1.22 package.

yes, you are right, debian buster (freezed) have 1.20, and we don't have mate 1.22 packages in debian by the moment

flexiondotorg:
I am +1 for merging this, it is a significant improvement and resolves a long standing problem.

flexiondotorg:
@sc0w FYI - I've pushed all git commits to Debian Salsa for MATE 1.22 but they've not been uploaded yet due to the Buster freeze. I'm sure MATE 1.22 will be uploaded soon after Buster is released.

raveit65:
> debian have compiz reloaded 0.8.16.1 in buster (testing freezed / future stable) and sid

And debian can update to 0.8.16.2 when released or adding those 4 compiz patches to existing package.

Should be possible for  such a smart and professional distro like debian?

PR #521, opened 2019-06-14
 bump LT_VERSION_INFO (soname version)

raveit65:
See docs https://www.sourceware.org/autobook/autobook/autobook_61.html#Library-Versioning

This is needed because current libmarco-private is incompatible to previous version.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,7 @@ dnl **************************************************************************
 dnl Library version information
 dnl **************************************************************************
 
-m4_define([lt_current], [1])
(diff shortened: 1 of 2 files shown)


lukefromdc:
Is anyone else going to review this or should we merge it now?

raveit65:
Normally a rebuild of marco and mate-control-center is needed if you use a package-manager for installing your packages.

lukefromdc:
I am guessing we will need another point release (1.23.2) for this as well

PR #527, opened 2019-07-09
switch to use meson build system

yetist:
Now meson build just works, please do more test.


lukefromdc:
Merged as this works, only adds files (for meson building) so can't break autotools, and all the checks passed

yetist:
Please don't forget to test the configure options, and I use 'disable-' as the option name like autotools for them, but I am not sure  use '-D disable-FEATURE' or '-D FEATURE' will be better.

PR #528, opened 2019-07-10
window: Update allowed action hints

vkareh:
When changing window state, we want to change the allowed action hints
so that other applications, mainly the taskbar, can disable menu entries
that do not make much visual sense. For example, unmaximizing
a minimized window: even though this operation is possible, it causes
user confusion as there is no visibility until the user unminimizes it.

Fixes https://github.com/mate-desktop/mate-panel/issues/860

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -66,7 +66,7 @@ static void     recalc_window_type        (MetaWindow     *window);
 static void     set_wm_state              (MetaWindow     *window,
                                            int             state);
 static void     set_net_wm_state          (MetaWindow     *window);
-
+static void     set_allowed_actions_hint  (MetaWindow     *window);
 static void     send_configure_notify     (MetaWindow     *window);
 static gboolean process_property_notify   (MetaWindow     *window,
                                            XPropertyEvent *event);
@@ -2469,6 +2469,8 @@ meta_window_minimize (MetaWindow  *window)
                                      queue_calc_showing_func,
                                      NULL);
 
+      set_allowed_actions_hint (window);
+
       if (window->has_focus)
         {
           meta_topic (META_DEBUG_FOCUS,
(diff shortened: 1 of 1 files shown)


vkareh:
### Unminimized/Unmaximized
unminimized-unmaximized

### Unminimized/Maximized
unminimized-maximized

### Minimized/Unmaximized
minimized-unmaximized

### Minimized/Maximized
minimized-maximized

sc0w:
I can't see in the code changes, the function `set_allowed_actions_hint`, is part of other library?

StealthStalker:
**_"For example, unmaximizing
a minimized window: even though this operation is possible, it causes
user confusion as there is no visibility until the user unminimizes it."_**

Thanks for fixing this guys, I can't follow this github site well at all so forgive my ignorance.  The above quote don't make sense to me though.  **"unmaximizing
a minimized window"**  How do you unmaximize a minimized window LOL.

I thought I posted a confirmation, and if so would someone point me to it.  I don't see it at my original "Issue" thread.  I did try this in Linux Mint Mate and it worked well.  What was weird is that I tried it another time prior and it was not working.  I'm sure it was the same iso, but when lukefromdc posted the fix I tried it and it was fixed...weird.

vkareh:
@StealthStalker - think of it this way:

1. You open a window and maximize it: maximized=true, visible=true
2. You minimize the window: maximized=true, visible=false
3. You unmaximize the window: maximized=false, visible=false
4. You unminimize the window: maximized=false, visible=true

The end result is that you end up with a window that's no longer maximized, so when you restore/unminimize the window, it will not restore itself to a maximized state.

So it's all correct if you think of it as window state flags, but because the operation is invisible to the user, it makes no sense to track it. Hence this patch.

StealthStalker:
I see, I'm certainly just a user and don't understand all the jargon but that makes sense as another way of looking at it, from a developers point of view.  Thanks SS!  I just thrilled that it works now!

PR #530, opened 2019-07-17
fix dark window decorations on apps which prefer dark theme

### LINKED ISSUE #105: support metacity-theme-3.xml

monsta:
This should completely fix #105, so that apps which prefer dark theme will also have dark window decoration if the theme supports that. If the theme doesn't support that, everything should work as before.

How to test:

1. Use some theme which has light and dark variants. Light variant should support dark theming for apps that prefer the dark theme (they set `gtk-application-prefer-dark-theme` in `GtkSettings`). I've used Arc theme.

2. Run some app which has the setting to prefer the dark theme. I've used gnome-terminal. Set the dark theme preference in `Edit > Preferences > General > Theme variant: Dark`.

Without this PR:

dark1

With this PR:

dark2

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -264,7 +264,7 @@ meta_window_new_with_attrs (MetaDisplay       *display,
   gulong existing_wm_state;
   gulong event_mask;
   MetaMoveResizeFlags flags;
-#define N_INITIAL_PROPS 19
+#define N_INITIAL_PROPS 20
   Atom initial_props[N_INITIAL_PROPS];
   int i;
   gboolean has_shape;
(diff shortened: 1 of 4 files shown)


monsta:
Ok, this needs more testing

monsta:
In Metacity it was apparently fixed by https://github.com/GNOME/metacity/commit/ba8500663457ad9f18ebfdf405162c2cb5caf88f, but that doesn't fix it here.

monsta:
Force-pushed with additional fixes. Now window decorations should be properly updated when changing themes.

monsta:
@rbuj: hmm... I reproduced the failed test (`GTK_THEME=Arc:dark marco-window-demo`) while my theme was set to Menta, and demo's window decorations stayed at Menta. The same happens with Metacity 3.30.1. This environment variable won't change window decoration theme...

muktupavels:
Why should that work?

monsta:
I think it shouldn't, I just replied about the failed test.

PR #535, opened 2019-08-25
window.c: remove unused variable

shlyakpavel:
Why would we need this var?


lukefromdc:
Could you change the commit messabe though, "cleanup window.c" is nowhere near as descriptive as say "window.c: remove unused variable"

shlyakpavel:
@lukefromdc no problem

PR #536, opened 2019-08-25
 display.c: add missing "break"

shlyakpavel:
I'm not 100% sure, but I feel like `break` is missing here

### DIFF
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -3809,7 +3809,7 @@ meta_display_begin_grab_op (MetaDisplay *display,
       meta_screen_ensure_tab_popup (screen,
                                     META_TAB_LIST_GROUP,
                                     META_TAB_SHOW_INSTANTLY);
-
+      break;
     case META_GRAB_OP_KEYBOARD_WORKSPACE_SWITCHING:
     case META_GRAB_OP_KEYBOARD_WORKSPACE_MOVING:
       meta_screen_ensure_workspace_popup (screen);


lukefromdc:
Codewise it certainly looks missing, presumably this would cause whatever triggers
` META_GRAB_OP_KEYBOARD_ESCAPING_GROUP:` to bring up both the tab and workspace popups simultaniously.

How exactly should this be tested, I don't know exactly what triggers that condition?

raveit65:
@shlyakpavel 
There is a open question from @flexiondotorg 
I would like to hear an answer from you.

raveit65:
@vkareh 
What do you think?

shlyakpavel:
> @shlyakpavel
> There is a open question from @flexiondotorg

Well
>How exactly should this be tested, I don't know exactly what triggers that condition?

I do not know

sc0w:
the new `break` line changes the behavior, and I don't know if it was missing by mistake, or intentionally

we have that code since the beginning of marco

shlyakpavel:
> the new `break` line changes the behavior, and I don't know if it was missing by mistake, or intentionally
> 
> we have that code since the beginning of marco

I hope you can find someone familiar with the code logic. Of course it changes the behavior and I really believe `break` was intended, but missed.

sc0w:
> 
>     Find a way to test his
>     Close the PR and forget it
>     Find a person who understands this

@shlyakpavel you can apply the same ways to yourself

shlyakpavel:
> Test:

Thanks a lot! 🥇

raveit65:
It is already fixed in metacity since a long time by their developers.
https://gitlab.gnome.org/GNOME/metacity/blob/master/src/core/display.c#L3711
So this looks ready to go.

Issue #540, opened 2019-09-02
[REG 1.22.1->1.22.2] No longer possible to go "exclusive" full screen


zaps166:
#### Expected behaviour
- disable compositor (or use compositor which doesn't redirect full screen windows),
- only one screen is connected,
- fullscreen OpenGL or Vulkan apps runs in page flipping mode,

#### Actual behaviour
- disable compositor (or use compositor which doesn't redirect full screen windows),
- only one screen is connected,
- fullscreen OpenGL or Vulkan apps runs in blit mode (it is slower and can cause screen tearing even if v-sync is enabled),

#### Steps to reproduce the behaviour
- disable compositor,
- if you use NVIDIA proprietary drivers, open `nvidia-settings`->"OpenGL settings" and check "Enable Graphics API Visual Indicator" - this allows you to easily check the "Flip" mode at runtime,
- be sure that you have single only one connected,
- run OpenGL or Vulkan application on fullscreen (e.g. game or video player),
- notice that NVIDIA Visual Indicator still displays "Blit" instead of "Flip" (in left-top corner). It is difficult to observe w/o NVIDIA indicator, but you can see lower performance or higher power consumption or sometimes screen tearing even is v-sync is enabled.

#### MATE general version
1.22.1

#### Package version
community/marco 1.22.2-1 (mate)

#### Linux Distribution
Manjaro Linux

#### Link to downstream report of your Distribution

---

Reverting commit fa32fffe0f6b0191cd17ce9ab19f89c4082497cf fixes the issue. AFAIK transparent windows (even if alpha is 0xff) won't allow flipping mode.

---

You can try (full screen) on mpv, QMPlay2, glxgears, [Vulkan examples](https://github.com/SaschaWillems/Vulkan), some Steam games like RoTR (but not all games), etc...

`glxgears -fullscreen` works, but running `glxgears` and then going full screen by a key shortcut does not.

lukefromdc:
Do you know if this is Nvidia-only or occurs with other drivers? You should be able test on your hardware with Nouveau or even llvmpipe to find out.

zaps166:
It is driver Independent, tested also on Mesa AMD driver, compositor disabled:
- marco 1.22.1 - screen tearing not observed,
- marco 1.22.2 - screen tearing observed,

[...]

PR #545, opened 2019-09-06
Fix use of RBGA visual in frame.c when compositing is not in use

lukefromdc:
Use the window's visual in all cases, fix problems with drivers forcing use of blit rather than pageflip mode when comppositing is not used or window is unredirected

Fix https://github.com/mate-desktop/marco/issues/540

Apply https://gitlab.gnome.org/GNOME/metacity/commit/5863176a2bd659c8d9a3d1c7b023a27c1a8c0aa5

### DIFF
--- a/src/core/frame.c
+++ b/src/core/frame.c
@@ -62,10 +62,7 @@ void
 meta_window_ensure_frame (MetaWindow *window)
 {
   MetaFrame *frame;
-  MetaScreen *screen;
   XSetWindowAttributes attrs;
-  XVisualInfo visual_info;
-  Visual *visual;
   int status;
 
   if (window->frame)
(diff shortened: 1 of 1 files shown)


muktupavels:
Artifact you see most likely is not visible on master before this was merged...

With ARGB window frame there is no black background (or at least should not have), but now with RGB frame is black if nothing paints in invisible area, but it definitely is not transparent. 

```
      if (visible_region)
        visible = cairo_region_to_xserver_region (xdisplay, visible_region);
```
Comment out this part in `border_size()` and compare what you get with argb and rgb frames.

My guess is that when you see "artifact" `meta_window_get_frame_bounds` returns `NULL` and compositor draws full window including invisible area.

PR #550, opened 2019-09-22
window decoration themes: Add Atlanta, Esco, Gorilla, Motif, Raleigh

rbuj:
### Atlanta

Screenshot at 2019-09-22 15-14-44

### Esco

Screenshot at 2019-09-22 15-55-55

### Gorilla

Screenshot at 2019-09-22 15-38-17

### Motif

Screenshot at 2019-09-22 15-15-35

### Raleigh
Screenshot at 2019-09-22 15-15-59


raveit65:
I recall that Atlanta theme was removed some years ago, i think it was in mate-themes.
And all those themes are metacity-theme-1.xml.
I would prefer to add metacity-theme-3.xml versions in 2019.

lukefromdc:
I just took a quick look at this (no time at the moment for a full build), and verified that there should be no conflicts with files currently installed by GNOME or Cinnamon. I have them both installed. Raleigh on my box prior to this has only a gtk-2.0 theme, so no conflict there, and the others are not present at all without applying this. Thus marco won't conflict with metacity or anything else I have installed by adding these back.

monsta:
Should we really ship more window themes that don't have a GTK+ (and icons) counterpart...?

vkareh:
> Should we really ship more window themes that don't have a GTK+ (and icons) counterpart...?

They don't need GTK+ or icons, they are just window decorations. The idea is that users can create their own mix-and-match theme using these components. I think this fits in well with the overall idea of MATE, in my opinion.

raveit65:
Btw. adding `metacity-theme-3.xml` to our existing themes is definitely a open task.

vkareh:
I'd say let's merge this. Since all current marco themes are `metacity-theme-1`, there is no inconsistency. Whether we upgrade them now or later is the same amount of work, so let's merge this and work on the theme upgrade separately. Thoughts?

flexiondotorg:
I agree. Let merge this and address the update to `metacity-theme-3` separately

PR #551, opened 2019-09-23
eOS theme: make transparent the button-box outer region

rbuj:
### Before

Captura de Pantalla 2019-09-23 a les 12 46 44

### After

Captura de Pantalla 2019-09-23 a les 12 47 41


rbuj:
@vkareh  I think it's an up scaling issue, because I didn't see any separation in a standard DPI resolution (no HiDPI) and I saw the separation using the magnifier of the gtk-inspector.

Captura de Pantalla 2019-09-23 a les 21 22 52

vkareh:
Interesting! Thanks for catching that, I'll test at regular scaling tomorrow in case the issue is on my side. But... Can it be fixed for HiDPI?

lukefromdc:
Same issue with standalone buttons in eOS from master also appears in compiz-reloaded and also is fixed by this (not surprisingly), though the tiny unscaled buttons are almost unusable on compiz-reloaded for now.

PR #557, opened 2019-09-26
Use WARN_CFLAGS which are only set with --enable-compiler-warnings

rbuj:
It removes --enable-more-warnings, since it is recommended to use
--enable-compile-warnings=maximum


raveit65:
@rbuj 
Can we use MATE_COMPILE_WARNINGS([yes]) and add compiler warnings in summary table, please?
I will also update travis config here.

raveit65:
Only updated mate-common version in travis config.

PR #568, opened 2019-10-28
window: Add optional tile size cycling

vkareh:
Adding a new option to allow tile size cycling. When enabled, using the
keyboard shortcut for tiling multiple times in a row cycles the window
through different sizes (1/2 -> 1/3 -> 1/4 -> 3/4 -> 2/3 -> Untiled).

### DIFF
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -3161,15 +3161,63 @@ handle_toggle_tiled (MetaDisplay *display,
                      MetaKeyBinding *binding)
 {
   MetaTileMode mode = binding->handler->data;
+  MetaTileCycle next_cycle;
 
-  if (mode == window->tile_mode && META_WINDOW_TILED(window))
+  /* Check if we're cycling through tile ratios */
+  if (meta_prefs_get_allow_tile_cycling ())
+    {
+      if (window->tile_mode != mode)
+        {
+          /* If changing tile mode, reset to first tile size */
+          next_cycle = META_TILE_CYCLE_50;
+        }
+      else
+        {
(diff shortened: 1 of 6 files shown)


raveit65:
Amazing feature :)
I think first tiling windows position should be `META_TILE_CYCLE_50`, or not?
Otherwise it is confusing when you expect quarter-tililing-left-down but window size is bigger.
Order should be `META_TILE_CYCLE_50` --> `META_TILE_CYCLE_66` --> `META_TILE_CYCLE_33` --> `META_TILE_CYCLE_NONE`
or
 `META_TILE_CYCLE_50` --> `META_TILE_CYCLE_33` --> `META_TILE_CYCLE_66` --> `META_TILE_CYCLE_NONE`.

raveit65:
I forgot.......
It should be the default, IHMO :)

raveit65:
After using it the first hour today it is clear that first window size when starting the cycle have to be `META_TILE_CYCLE_50` or `1/2` size, everything else is confusing ;) 
But it works very nice.

vkareh:
Thanks for testing and for your suggestions. I've updated it so that it always starts with 50% of the screen, then goes smaller from there and back to larger, and it feels a lot better. Also I've added 2 extra sizes.

vkareh:
> I forgot.......
> It should be the default, IHMO :)

Done

vkareh:
@lukefromdc thanks for testing. The tile size choices are also influenced by Ubuntu's Compiz, which has the cycle: 50->33->25->100->75->66. In testing different configurations and orders with @raveit65, we decided to drop the 100 (we have maximize for that), but keep the others, and add the Untiled option to the list (since it follows marco's current behavior, where keyboard tiling toggles between 50 and Untiled).

lukefromdc:
Oh yeah: I was comparing to compiz-reloaded, which I have and work on

Issue #560, opened 2019-10-04
marco window foreground/background behaviour broken when switching workspaces


pasikarkkainen:
With the default Mate/Marco settings "Enable software compositing window manager" -> "enabled" Marco is very problematic/broken when switching workspaces. When you have many windows open on a workspace, and you switch between different workspaces (either with a mouse from the workspace switcher applet, or with a keyboard shortcut), it seems some random window gets the focus but the focused window is not on the foreground, instead it's somewhere on the background with all the other (or at least most of the other) windows in front of it.. 

1) you can't see the focused window properly, because it's in the background, all the other / most of the other windows are being in front of it / "hiding" it.
2) you can't click/interact with what you see on the desktop/workspace, because all the input goes to that weird/broken focused window in the background, not to that window which you see and try to interact with..

Restoring the focused window to foreground is very tricky, because all the focus goes to this broken focused window in the background, so you can't easily minimize the other windows if/when they're overlapping, because like said the focus goes to that weird focused window in the background instead of the window you try to minimize.. easiest way to get the focused window to foreground is to click the "hide all windows"-button, and then find the wanted window to show it.

Disabling the "Enable software compositing window manager" setting fixes the problematic behaviour.

This feels/seems like exactly the same issue as a previous marco issue ( https://github.com/mate-desktop/marco/issues/251 ). copy/paste from the previous issue:

"windows is not brought to the front, it stays in the back. So, it looks like I'm clicking on the correct window, but I'm clicking in one in the background."

#### Expected behaviour
Focused window is on the foreground after switching workspaces.

#### Actual behaviour
Focused window goes to background, with many (if not all?) windows in front of it.. so you can't see the focused window, except all the input goes to that weird/broken window in the background.

#### Steps to reproduce the behaviour
Use mate/marco with the default settings (software compositing window manager -> enabled), have many windows open on each workspace, and switch between workspaces a few times.

#### MATE general version
mate 1.22

#### Package version
marco-1.22.3-1.fc30.x86_64

#### Linux Distribution
Fedora 30.

I'm seeing this same broken behaviour on multiple systems: on a single-display laptop, and also on a multi-display workstation.

raveit65:
Does this happens with marco-1.22.1-1.fc30 ?
https://koji.fedoraproject.org/koji/buildinfo?buildID=1256952
Which GPU and driver are you using?

Honestly, i can't reproduce it with nvidia GPU and proprietary nvidia driver.
[...]

PR #570, opened 2019-12-02
Revert "compositor: fix possible crash closing/destroying window"

vkareh:
This reverts commit df56628979abc17474b001c58c28c6f3637ada9d.

It causes an issue where windows remain in the background even when
being drawn in the foreground.


raveit65:
Probably a fix for https://github.com/mate-desktop/marco/issues/560

PR #572, opened 2019-12-05
theme.c: Fix window control hidpi rendering for all themes.

mtwebster:
By scaling the pixbuf when loading, existing assets can be used.

meta_theme_load_image() is only run once for each image when the theme loads, so the cost of loading them twice is relatively minor.

Examples (in hidpi):
Screenshot from 2019-12-05 12-13-18

Screenshot from 2019-12-05 12-09-52
(this is with or without https://github.com/mate-desktop/mate-themes/commit/8f56d6be5a2e3116699703da9f103a964ee7222d)

### DIFF
--- a/src/ui/theme.c
+++ b/src/ui/theme.c
@@ -4134,8 +4134,8 @@ meta_draw_op_draw_with_env (const MetaDrawOp    *op,
 
         if (op->data.image.pixbuf)
           {
-            env->object_width = gdk_pixbuf_get_width (op->data.image.pixbuf);
-            env->object_height = gdk_pixbuf_get_height (op->data.image.pixbuf);
+            env->object_width = gdk_pixbuf_get_width (op->data.image.pixbuf) / scale;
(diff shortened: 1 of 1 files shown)


mtwebster:
Thanks, I honestly never noticed that function before - always learning.  Updated.

vkareh:
Could you squash the commits, please? The second commit message is unnecessary, I think

mtwebster:
Done

vkareh:
Thanks a bunch :)

raveit65:
@vkareh 
Should we consider this as bugfix and cherry-pick to 1.22?

vkareh:
Good point. This is a bugfix, we should cherry pick

raveit65:
We are faster as you could know that you have a wish :)
https://github.com/mate-desktop/marco/commit/75c6e6c3682bc8ac5ef4803a6670b2cc57ed5c4f

PR #577, opened 2020-01-11
Xprops fix memory leak.

rcaridade145:
Backports a fix for a memory leak fixed on Metacity but not on Marco.
This brings down normal usage on my machine from 256 Mb to 180 something

Origin commit :
https://gitlab.gnome.org/GNOME/metacity/commit/c87f73f3b4413720a2f3e6a672826d3fec7f77a9
"
XmbTextPropertyToTextList documentation says that XFreeStringList
should be used to free the storage for the list and its contents.
"


raveit65:
Seems Ok when it is from @muktupavels , i am just testing.

Issue #316, opened 2017-04-26
Chrome/Chromium stuck maximised on secondary screen with Marco


Kallestofeles:
Hello,

A strange issue which happens only with Marco WM.
When maximising the Chrome/Chromium window on a secondary monitor, I can no longer control the window. Cannot minimise it, resize it, nor move it. I can only close the window.
Best to explain the behaviour with a video:
[https://goo.gl/photos/sLFUgniFs5niVyC66](url=https://goo.gl/photos/sLFUgniFs5niVyC66)
In the video, the Picture in Picture screen is the primary screen1 and the big screen on the background is the secondary screen2. Resolutions are the same and the issue only occurs while using Marco WM. (tested with xfwm, compiz0.9, mutter, all work fine except Marco)

Sidenote: when viewing the video, please don't mind the background picture, it means no offence.

If there is any info or details I could provide, please let me know, I'm happy to help as I wish this to get resolved.

OS: Arch Linux
Kernel: x86_64 Linux 4.10.11-1-ARCH
DE: MATE 1.18.0
WM: Metacity (Marco)
marco 1.18.0-1

Thank you!

Best regards,
Kalle

Kallestofeles:
Update:
Also happens with Steam.
Looks like it's related to non-native window decorations and Marco.

muktupavels:
Did you try Metacity? Does it work? If you test with it, please tell also version that you used for testing.

Kallestofeles:
Hi muktupavels,

Thank you for responding!
Tried now with pure Metacity and the issue persists with that as well. So if I'm right, then Marco is built upon metacity but the issue is with that. Hmmh.
`[kallestofeles@Shoemaster ~]$ metacity --version
metacity 3.24.1
`

muktupavels:
I think I was able to reproduce this bug / problem. Can you please open upstream bug for metacity:
https://bugzilla.gnome.org/enter_bug.cgi?product=metacity

Kallestofeles:
Created the case with Gnome team:
https://bugzilla.gnome.org/show_bug.cgi?id=781946
Issue can be resolved by issuing "--no-force-fullscreen" option when starting Marco/Metacity. This should be the default as I cannot see any drawbacks so far... I'll update if I find anything.

deltragon:
I have encountered the same issue, but I've found a workaround for Chrome: right clicking Chrome's window decorator gives you an option to "Use system title bar and borders", which resolves it (even if it looks a bit ugly).

[...]

PR #581, opened 2020-01-25
Dual monitor windows stuck fullscreen fix

rcaridade145:
This PR backports the commit https://gitlab.gnome.org/GNOME/metacity/commit/1fafd279006ece8cf664fd777143cdfafbefad6d on Metacity to Marco.
It should fix issue #316 .
I don't have multi monitors so i cannot test.
It supersedes the troubled PR i did :-1: 

Bug 781946 - Non-native decorated windows stuck maximised on secondary screen with Marco/Metacity

### DIFF
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -447,29 +447,6 @@ setup_constraint_info (ConstraintInfo      *info,
     meta_workspace_get_onxinerama_region (cur_workspace,
                                           xinerama_info->number);
 
-  /* Workaround braindead legacy apps that don't know how to
-   * fullscreen themselves properly - don't get fooled by
-   * windows which hide their titlebar when maximized or which are
-   * client decorated; that's not the same as fullscreen, even
-   * if there are no struts making the workarea smaller than
-   * the monitor.
-   */
-
-  if (meta_prefs_get_force_fullscreen() &&
-      (window->decorated || !meta_window_is_client_decorated (window)) &&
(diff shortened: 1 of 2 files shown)


raveit65:
The PR fixes the problem with chrome/chromium if "use system titlebar and border" option is unchecked, but i like to test it a bit more.

lukefromdc:
I cannot test the one-off CSD windows of chromium as I don't have that installed, but I found that ALL conventional windows (both CSD and SSD) try to maximize across both monitors with one 4K and one 1080p monitor (both at window-scaling=2 as screwing with xrandr to make them different is too slow) try to maximize across both monitors with ugly results. 

About to test this and see if it has any effect

vkareh:
@mate-desktop/core-team - sounds like a candidate for 1.22 cherry-picking. Thoughts?

monsta:
"Workaround braindead legacy apps that don't know how to fullscreen themselves properly" - looks like Chromium is a truly classic app? :grin: 

I've picked it to 1.22 (897a11c196081a31b3cd6035270da28039cb0e52).
I guess we can close https://github.com/mate-desktop/marco/issues/316 now?

raveit65:
> I've picked it to 1.22 (897a11c).

You want to make new releases? Otherwise it is useless ;)

raveit65:
> I guess we can close #316 now?

Why always asking, just do it :)

PR #582, opened 2020-01-31
Fix some deprecation warnings

rcaridade145:
Based on

https://gitlab.gnome.org/GNOME/metacity/commit/5aecb8336c6c90a9ad1bb46295ca60624f5a44aa
https://gitlab.gnome.org/GNOME/metacity/commit/1dcde194923c2f6791de0c6b5d6bd6c3b66b80da
https://gitlab.gnome.org/GNOME/metacity/commit/359f60a6aee5f1338a5edd9f4d84557126c6d46e

### DIFF
--- a/src/ui/frames.c
+++ b/src/ui/frames.c
@@ -2261,28 +2261,6 @@ meta_frames_destroy_event           (GtkWidget           *widget,
   return TRUE;
 }
 
-static void
-setup_bg_cr (cairo_t *cr, GdkWindow *window, int x_offset, int y_offset)
-{
-  GdkWindow *parent = gdk_window_get_parent (window);
-  cairo_pattern_t *bg_pattern;
-
-  bg_pattern = gdk_window_get_background_pattern (window);
-  if (bg_pattern == NULL && parent)
-    {
-      gint window_x, window_y;
-
-      gdk_window_get_position (window, &window_x, &window_y);
-      setup_bg_cr (cr, parent, x_offset + window_x, y_offset + window_y);
-    }
-  else if (bg_pattern)
-    {
-      cairo_translate (cr, - x_offset, - y_offset);
-      cairo_set_source (cr, bg_pattern);
-      cairo_translate (cr, x_offset, y_offset);
-    }
-}
-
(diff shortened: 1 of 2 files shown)


rcaridade145:
That work was based on metacity and a particular commit. I can try and restore that functionality as suggested here https://developer.gnome.org/gtk3/stable/GtkImageMenuItem.html

raveit65:
No, please do not Icons from menus or buttons.
We are MATE and not Gnome......

raveit65:
Why not using 3 commits with meaningful names like gnome does?
1. menu: replacing GtkStock deprecation
2. frames: don't bother painting the background

    GDK (and also GTK+!) do this for us already.
3. theme: libmetacity: avoid deprecated g_memmove

rcaridade145:
I've been working on this and got to the problem discussed here [https://stackoverflow.com/questions/48452717/how-to-replace-the-deprecated-gtk3-gtkimagemenuitem](url) 
left_align_small

This being said i reverted to the old behavior .

rcaridade145:
> Why not using 3 commits with meaningful names like gnome does?
> 
>     1. menu: replacing GtkStock deprecation
> 
>     2. frames: don't bother painting the background
>        GDK (and also GTK+!) do this for us already.
> 
>     3. theme: libmetacity: avoid deprecated g_memmove

Sorry will do that on the next PR's i've got on the pipeline.

src/ui/menu.c
```diff
@@ -60,6 +60,7 @@ struct _MenuData {
 
 static void activate_cb(GtkWidget* menuitem, gpointer data);
 
+
```
vkareh: This change belongs to a commit that's no longer part of this. Please remove

src/ui/menu.c
```diff
@@ -311,6 +312,7 @@ static GtkWidget* menu_item_new(MenuItem* menuitem, int workspace_id)
 	return mi;
 }
 
+
```
vkareh: Same here...

raveit65:
Thanks, for updating PR in single commits.
I just tested it and it runs fine.

PR #584, opened 2020-02-03
remove warning: 'GTimeVal' is deprecated

rbuj:


rbuj:
@vkareh You're right [g_get_real_time](https://developer.gnome.org/glib/stable/glib-Date-and-Time-Functions.html#g-get-real-time) is available since GLib 2.28.

rbuj:
@vkareh done

PR #597, opened 2020-02-23
meson: improve

yetist:
Drop the enable from `--enable-foo` boolean options; Meson has boolean values, so `-Denable-foo=true` would read as redundant, and `-Denable-foo=false` would read as contradictory. Use `-Dfoo=true` or `-Dfoo=false` instead

https://wiki.gnome.org/Initiatives/GnomeGoals/MesonPorting

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -191,8 +191,7 @@ xpresent_dep = dependency('xpresent', required: false)
 xinerama_dep = dependency('xinerama', required: false)
 sm_dep = dependency('sm', required: false)
 
-verbose_mode = not get_option('disable-verbose-mode')
-if verbose_mode
+if get_option('verbose-mode')
   config_h.set('WITH_VERBOSE_MODE', 1, description: 'Build with verbose mode support')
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -281,14 +277,13 @@ if libm.found()
   marco_deps += [libm]
 endif
 
-build_sm = not get_option('disable-sm')
-if build_sm and not sm_dep.found()
+build_sm = false
```
raveit65: Why `build_sm = false` ?
In meson_options.txt the option `sm` is enabled as default, or not?

yetist: `build_sm` is a variable used to display information at the end of the configuration:
```
  ' Session management:       @0@'.format(build_sm),
```

`build_sm` is false by default and is set to true only when dependencies are found:
```
if get_option('sm') and sm_dep.found() and ice_dep.found()
  build_sm = true
endif
```

raveit65: Good.

PR #625, opened 2020-05-30
compositor-xrender.c: keep tooltips visible in hidpi.

mtwebster:
Make sure tooltips are visible when the source widget is close to the edge of the work area. widget is close to the edge of the work area.

See inline comments - as of 3.24 tooltip positioning is handled
differently, and under certain conditions in hidpi, tooltips for the
bottom panel applets can end up off the bottom of the work area.

To reproduce, in hidpi, set the bottom panel to approximately 30px
tall and try to activate a tooltip for an applet on that panel.

- I'm not too familiar with this compositor, it's possible there's a better place for this sort of fix.
- I had assumed that since panels can be added to any monitor that I could get the precise work area for any monitor (not just for the primary) but that doesn't seem to be possible - tooltips will overlap the panel on a second monitor.

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -41,6 +41,8 @@
 
 #include "display.h"
 #include "../core/display-private.h"
+#include "../core/screen-private.h"
+#include "../core/workspace.h"
 #include "screen.h"
 #include "frame.h"
 #include "errors.h"
@@ -48,6 +50,7 @@
 #include "compositor-private.h"
 #include "compositor-xrender.h"
 #include "xprops.h"
+#include "core.h"
 #include <X11/Xatom.h>
 #include <X11/extensions/shape.h>
 #include <X11/extensions/Xcomposite.h>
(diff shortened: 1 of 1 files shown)


muktupavels:
Sound like something that should be fixed in GTK...

lukefromdc (approved):
Builds and runs fine, and fixes 
https://github.com/mate-desktop/mate-panel/issues/1180
Tooltips render slightly high on a 4K monitor with a smaller secondary monitor attached but turned off, but that's a hell of a lot better than rendering them too low.

lukefromdc:
Any updates on this?

raveit65:
@vkareh 
Can you please do a review?
It runs fine since half a year at fedora stable all branches.
@lukefromdc 
No idea why Linuxmint head and Mate founder (Clement Lefebvre) do not  review his own developer from LM.
Any way, i am fine with merging it, because i didn't got any bugreport at RedHat Bugzilla with that commit.

lukefromdc:
Should I wait to see if that review comes in or proceed to merge this?

PR #633, opened 2020-07-28
WIP Fix some warnings about float conversion (-Wfloat-conversion)

rbuj:
Test: marco.make.log (master), marco/make.log (PR)
```
$ CFLAGS="-O3 -fomit-frame-pointer -malign-double -fstrict-aliasing -ffast-math -Wconversion -Wunused-macros -Wunused-parameter" ./autogen.sh --prefix=/usr --enable-compile-warnings=maximum && make &> make.log && sudo make install
```
Comparing make logs
```
$ diff <(./warnings-file.sh marco.make.log) <(./warnings-file.sh marco/make.log)
3c3
< -Wconversion 978
---
> -Wconversion 951
5c5
< -Wfloat-conversion 97
---
> -Wfloat-conversion 18
```
warnings-file.sh
```
FILE=$1
for warning in $(grep warning $FILE | grep -Po '\[\-W.*\]' | sed 's/[][]//g' | sort -u)
do
 type=${warning#"-W"}
 num=$(echo "grep $type $FILE | wc -l" | sh)
 echo "$warning $num"
done
```
Testing scaled surfaces: save scaled surfaces to png files "/home/robert/SC_TIME " + TIMESTAMP + ".png"
- Run mate-appearance-properties
- Select Theme tab
```C
static cairo_surface_t *
scale_surface (cairo_surface_t *surface,
               gdouble          old_width,
               gdouble          old_height,
               gdouble          new_width,
               gdouble          new_height,
               gboolean         vertical_stripes,
               gboolean         horizontal_stripes)
```
```diff
diff --git a/src/ui/theme.c b/src/ui/theme.c
index be3e87a..6c7a055 100644
--- a/src/ui/theme.c
+++ b/src/ui/theme.c
@@ -140,6 +140,10 @@ scale_surface (cairo_surface_t *surface,
   cairo_pattern_set_extend (cairo_get_source (cr), CAIRO_EXTEND_PAD);
 
   cairo_paint (cr);
+  char *file = g_strdup_printf ("/home/robert/SC_TIME%lu.png", g_get_monotonic_time ());
+  cairo_surface_write_to_png (surface, file);
+  g_free (file);
+
   cairo_destroy (cr);
 
   return scaled;
```
Testing Alt+Tab thumbnails: save window screenshots to png files "/home/robert/SC_TIME " + TIMESTAMP + ".png"
- Press Alt+Tab
```C
static cairo_surface_t *
get_window_surface (MetaWindow *window)
```
```diff
diff --git a/src/core/screen.c b/src/core/screen.c
index 1218232..a821be5 100644
diff --git a/src/core/screen.c b/src/core/screen.c
index 1218232..6a300ff 100644
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -1265,6 +1265,10 @@ get_window_surface (MetaWindow *window)
   cairo_destroy (cr);
   cairo_surface_destroy (surface);
 
+  char *file = g_strdup_printf ("/home/robert/SC_TIME%lu.png", g_get_monotonic_time ());
+  cairo_surface_write_to_png (scaled, file);
+  g_free (file);
+
   return scaled;
 }
 
```
benchmark: master vs PR (there is no significant difference)
```
$ marco-theme-viewer BlueMenta
```
### master

Screenshot at 2020-08-09 21-53-07

### PR

Screenshot at 2020-08-09 21-54-37

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -833,10 +833,10 @@ solid_picture (MetaDisplay *display,
       return None;
     }
 
-  c.alpha = a * 0xffff;
-  c.red = r * 0xffff;
-  c.green = g * 0xffff;
-  c.blue = b * 0xffff;
+  c.alpha = (guint16)(a * 65535 + 0.5);
+  c.red   = (guint16)(r * 65535 + 0.5);
+  c.green = (guint16)(g * 65535 + 0.5);
+  c.blue  = (guint16)(b * 65535 + 0.5);
 
   XRenderFillRectangle (xdisplay, PictOpSrc, picture, &c, 0, 0, 1, 1);
   XFreePixmap (xdisplay, pixmap);
--- a/src/core/place.c
+++ b/src/core/place.c
@@ -31,6 +31,7 @@
 #include "prefs.h"
 #include <gdk/gdk.h>
 #include <math.h>
+#include <float.h>
 #include <stdlib.h>
 
 typedef enum
@@ -46,8 +47,8 @@ northwestcmp (gconstpointer a, gconstpointer b)
 {
   MetaWindow *aw = (gpointer) a;
   MetaWindow *bw = (gpointer) b;
-  int from_origin_a;
-  int from_origin_b;
+  float from_origin_a;
+  float from_origin_b;
   int ax, ay, bx, by;
 
   /* we're interested in the frame position for cascading,
@@ -76,8 +77,8 @@ northwestcmp (gconstpointer a, gconstpointer b)
     }
 
   /* probably there's a fast good-enough-guess we could use here. */
-  from_origin_a = sqrt (ax * ax + ay * ay);
-  from_origin_b = sqrt (bx * bx + by * by);
+  from_origin_a = sqrtf ((float)(ax * ax + ay * ay));
+  from_origin_b = sqrtf ((float)(bx * bx + by * by));
 
   if (from_origin_a < from_origin_b)
     return -1;
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -1211,7 +1211,7 @@ meta_screen_update_cursor (MetaScreen *screen)
 }
 
 #define MAX_PREVIEW_SCREEN_FRACTION 0.33
-#define MAX_PREVIEW_SIZE 300
+#define MAX_PREVIEW_SIZE 300.0
 
 static cairo_surface_t *
 get_window_surface (MetaWindow *window)
@@ -1219,7 +1219,8 @@ get_window_surface (MetaWindow *window)
   cairo_surface_t *surface, *scaled;
   cairo_t *cr;
   const MetaXineramaScreenInfo *current;
-  int width, height, max_columns, max_size;
+  int width, height, max_columns;
+  double max_size;
   double ratio;
 
   surface = meta_compositor_get_window_surface (window->display->compositor, window);
@@ -1237,16 +1238,16 @@ get_window_surface (MetaWindow *window)
   if (width > height)
     {
       max_size = MIN (MAX_PREVIEW_SIZE, current->rect.width / max_columns * MAX_PREVIEW_SCREEN_FRACTION);
-      ratio = ((double) width) / max_size;
+      ratio = max_size / width;
       width = (int) max_size;
-      height = (int) (((double) height) / ratio);
+      height = (int) (height * ratio + 0.5);
     }
   else
     {
       max_size = MIN (MAX_PREVIEW_SIZE, current->rect.height / max_columns * MAX_PREVIEW_SCREEN_FRACTION);
-      ratio = ((double) height) / max_size;
+      ratio = max_size / height;
       height = (int) max_size;
-      width = (int) (((double) width) / ratio);
+      width = (int) (width * ratio + 0.5);
     }
 
   meta_error_trap_push (window->display);
@@ -1257,7 +1258,7 @@ get_window_surface (MetaWindow *window)
     return NULL;
 
   cr = cairo_create (scaled);
-  cairo_scale (cr, 1/ratio, 1/ratio);
+  cairo_scale (cr, ratio, ratio);
   cairo_set_source_surface (cr, surface, 0, 0);
   cairo_paint (cr);
 
--- a/src/core/constraints.c
+++ b/src/core/constraints.c
@@ -541,15 +541,15 @@ place_window_if_needed(MetaWindow     *window,
            */
           if (info->current.width >= info->work_area_xinerama.width)
             {
-              info->current.width = .75 * info->work_area_xinerama.width;
+              info->current.width = (int) (.75 * info->work_area_xinerama.width + 0.5);
               info->current.x = info->work_area_xinerama.x +
-                       .125 * info->work_area_xinerama.width;
+                       (int)(.125 * info->work_area_xinerama.width + 0.5);
             }
           if (info->current.height >= info->work_area_xinerama.height)
             {
-              info->current.height = .75 * info->work_area_xinerama.height;
+              info->current.height = (int)(.75 * info->work_area_xinerama.height + 0.5);
               info->current.y = info->work_area_xinerama.y +
-                       .083 * info->work_area_xinerama.height;
+                       (int)(.083 * info->work_area_xinerama.height + 0.5);
             }
 
           if (window->maximize_horizontally_after_placement ||
@@ -1293,14 +1293,12 @@ constrain_aspect_ratio (MetaWindow         *window,
     {
     case WestGravity:
     case EastGravity:
-      /* Yeah, I suck for doing implicit rounding -- sue me */
-      new_height = CLAMP (new_height, new_width / maxr,  new_width / minr);
+      new_height = CLAMP (new_height, (int)((new_width / maxr) + 0.5),  (int)((new_width / minr) + 0.5));
       break;
 
     case NorthGravity:
     case SouthGravity:
-      /* Yeah, I suck for doing implicit rounding -- sue me */
-      new_width  = CLAMP (new_width,  new_height * minr, new_height * maxr);
+      new_width  = CLAMP (new_width,  (int)((new_height * minr) + 0.5), (int)((new_height * maxr) + 0.5));
       break;
 
     case NorthWestGravity:
@@ -1312,8 +1310,8 @@ constrain_aspect_ratio (MetaWindow         *window,
     default:
       /* Find what width would correspond to new_height, and what height would
        * correspond to new_width */
-      alt_width  = CLAMP (new_width,  new_height * minr, new_height * maxr);
-      alt_height = CLAMP (new_height, new_width / maxr,  new_width / minr);
+      alt_width  = CLAMP (new_width,  (int)((new_height * minr) + 0.5), (int)((new_height * maxr) + 0.5));
+      alt_height = CLAMP (new_height, (int)((new_width / maxr) + 0.5),  (int)((new_width / minr) + 0.5));
 
       /* The line connecting the points (alt_width, new_height) and
        * (new_width, alt_height) provide a range of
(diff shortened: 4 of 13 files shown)


raveit65:
@rbuj 
We have a conflict...

rbuj:
@raveit65 done. I fixed the merge conflict.

src/compositor/compositor-xrender.c
```diff
@@ -833,10 +833,10 @@ solid_picture (MetaDisplay *display,
       return None;
     }
 
-  c.alpha = a * 0xffff;
-  c.red = r * 0xffff;
-  c.green = g * 0xffff;
-  c.blue = b * 0xffff;
+  c.alpha = (guint16)(a * 65535 + 0.5);
```
muktupavels, line 836: Why `+ 0.5` everywhere?

src/core/place.c
```diff
@@ -76,13 +77,16 @@ northwestcmp (gconstpointer a, gconstpointer b)
...
 
   /* probably there's a fast good-enough-guess we could use here. */
-  from_origin_a = sqrt (ax * ax + ay * ay);
-  from_origin_b = sqrt (bx * bx + by * by);
+  from_origin_a = sqrtf ((float)(ax * ax + ay * ay));
+  from_origin_b = sqrtf ((float)(bx * bx + by * by));
 
-  if (from_origin_a < from_origin_b)
+  /* from_origin_a < from_origin_b */
+  if (fabsf (from_origin_b - from_origin_a) >= FLT_EPSILON)
```
muktupavels, line 84: Why?

src/core/screen.c
```diff
@@ -1237,16 +1238,16 @@ get_window_surface (MetaWindow *window)
...
+      ratio = max_size / width;
       width = (int) max_size;
-      height = (int) (((double) height) / ratio);
+      height = (int) (height * ratio + 0.5);
     }
   else
     {
       max_size = MIN (MAX_PREVIEW_SIZE, current->rect.height / max_columns * MAX_PREVIEW_SCREEN_FRACTION);
-      ratio = ((double) height) / max_size;
+      ratio = max_size / height;
```
muktupavels, line 1248: Why?

rbuj: performance and accuracy

src/compositor/compositor-xrender.c
rbuj: It's the recommended and the generic way to convert floating point numbers to integers.

src/core/place.c
rbuj: This is how I'm used to compare the floating point numbers on digital signal processors, also in java applications, so I think it's also correct here. Or you mean the ends, whether they are open or closed.

muktupavels: I mean how this is related to warnings you are fixing in this MR?

src/core/screen.c
rbuj: A division is more expensive than a product when working with floating point numbers, without counting the latency the order usually is 2:1 on single precision (float), a little more if it's on double precision (double).

The single precision numbers (float) have 7 decimal digits` log10(2^24) ≈ 7.225 `, you can find more information about float accuracy at https://en.wikipedia.org/wiki/Single-precision_floating-point_format

muktupavels: Why are you adding unrelated changes in commit that fix build warning?

src/core/place.c
rbuj: it's to not go to work with integers

rbuj: If you think the comparison that was there before it's simply better, I have no problem leaving the comparison as it was before.

src/compositor/compositor-xrender.c
muktupavels: Who recommends that? Generic way?

src/core/screen.c
rbuj: IMHO, I think they are related but I may be wrong.

src/core/screen.c
```diff
@@ -1210,16 +1210,17 @@ meta_screen_update_cursor (MetaScreen *screen)
   XFreeCursor (screen->display->xdisplay, xcursor);
 }
 
-#define MAX_PREVIEW_SCREEN_FRACTION 0.33
-#define MAX_PREVIEW_SIZE 300
+#define MAX_PREVIEW_SCREEN_FRACTION 0.33f
```
rbuj, line 1213: It maybe it should be double...

src/compositor/compositor-xrender.c
rbuj: It recall that it's the JTC1/SC22/WG11, but I may be wrong.

rbuj: How would you do it? truncating?

rbuj: When you do a floating point conversion to an integer, you can represent it as a staircase on a graph, where each step is the integer value. You have to take into account the number of digits you have for accuracy (floating point number representation), and the range of the interval (number of steeps). If you have many step, the conversion error will be smaller or if you have a lot of accuracy (sampling theory).

Truncating means to remove the decimal part from a number. Truncating a scaling result (max_value * factor) means to remove the last step of the staircase, for this reason rounding is done (+ 0.5). 
```
integer_value = max_value * factor, 0 <= factor <= 1
```

rbuj: Anyway, I think explaining this doesn’t make much sense here, and I guess there are many others who can explain it better than me. :)

rbuj: Now that we talk about it, I don't know why glib developers always use double precision floating point number for graphics, penalizing devices with limited capabilities like on the raspberry pi platform. But I suppose we have to live with this choice ... I guess someone got it wrong and did it thinking it works just like going from a 32 to 64 bit architecture.

src/core/place.c
rbuj: done
```
diff --git a/src/core/place.c b/src/core/place.c
index fa5c8e0..40370e7 100644
--- a/src/core/place.c
+++ b/src/core/place.c
@@ -80,13 +80,10 @@ northwestcmp (gconstpointer a, gconstpointer b)
   from_origin_a = sqrtf ((float)(ax * ax + ay * ay));
   from_origin_b = sqrtf ((float)(bx * bx + by * by));
 
-  /* from_origin_a < from_origin_b */
-  if (fabsf (from_origin_b - from_origin_a) >= FLT_EPSILON)
+  if (from_origin_a < from_origin_b)
     return -1;
-  /* from_origin_a > from_origin_b */
-  else if (fabsf (from_origin_a - from_origin_b) >= FLT_EPSILON)
+  else if (from_origin_a > from_origin_b)
     return 1;
-  /* from_origin_a == from_origin_b */
   else
     return 0;
 }
```

rbuj: @muktupavels recommend us not using an epsilon for comparing two floating numbers.

rbuj:
@muktupavels We talked about this some time ago in another PR #476, I thought you participated in the conversation, but I was wrong. All I'm telling you is from memory:

At that moment of writing the other PR, I was focused on the performance after seeing an unusual comment on the code.

Doing it that way introduced another warning, From memory, do not convert directly the returned value of a function i.e. do the conversion using a variable, because of mismatch on datatypes, since round returned double and it was stored on a non double variable.

In other parts of the code I think I remember that they use the same principle, rounding up or down depending on whether the value is closer to one of the ends, i.e. using the round operation. I think it's done using macros to avoid the round function of math.h. What’s more, I think there is also a scaling macro, which cannot be reused by the data type, i.e. wide or sign, which is not much different from what this PR proposes.

src/compositor/compositor-xrender.c
muktupavels: I think change **without** explanation does not make any sense... Sorry, but this is first time that I hear that using `+ 0.5` is recommended way to convert float to int as generic way.

Adding `+ 0.5` everywhere is not needed to fix warning you are fixing with this MR. You are mixing warning fix with rounding behavior changes. What bug/issue are you trying to solve with this change?

In one comment you mention performance as reason and yet you make CPU do more work for no real reason.

src/core/screen.c
muktupavels: From MR title you are fixing `-Wfloat-conversion`, no? It might be okayish if you would have included needed description in commit message.

And if you speak about performance include numbers. You gain nothing here, it is not performance critical path...

Maybe check - https://softwareengineering.stackexchange.com/questions/80084/is-premature-optimization-really-the-root-of-all-evil.

src/compositor/compositor-xrender.c
rbuj: I have not invented anything new, color scaling function on mate-desktop/libmate-desktop/mate-colorsel.c 
```C
/* Conversion between 0->1 double and and guint16. See
 * scale_round() below for more general conversions
 */
#define SCALE(i) (i / 65535.)
#define UNSCALE(d) ((guint16)(d * 65535 + 0.5))
#define INTENSITY(r, g, b) ((r) * 0.30 + (g) * 0.59 + (b) * 0.11)
```

muktupavels:
You asked for review and you got one...

src/core/screen.c
rbuj: I don't know, but I see it more feasible to use the factor than its inverse.

rbuj:
I'm very grateful to you, I will work to include the improvements you have asked me.

rbuj:
@muktupavels I added the text below on the commit body:
```
- round on unscale (scale_factor * max_value + 0.5)
- do not require to invert float: cairo_scale (cr, 1/ratio, 1/ratio)
- use specific operations based on datatype: sqrtf for float
- no mixing single precision with double precision numbers
```

muktupavels:
@rbuj Now imagine that it was not your work but was done by somebody else. You have found that this commit has introduced regression. Does that message gives anything useful to understand why change was made? Did it fix something?

Adding 0.5 can easily introduce regression because original author might have relied on truncating. Obviously you will not notice any regression in `solid_picture()`... Getting 32768 instead of 32767 for 0.5 changes nothing. If you would have looked how that function is used than you would notice that it is used only in 3 places, one is only if debugging redraws. You could use `unsigned short` for r, g, b parameters... 

Also I still think this should be done in more smaller commits.

src/compositor/compositor-xrender.c
muktupavels: That was added as bug fix - https://bugzilla.gnome.org/show_bug.cgi?id=58120. What are you fixing here?

rbuj:
Ok, I'm goin to close this MR and I'll do small MRs, the 1st #638

src/compositor/compositor-xrender.c
rbuj: Yes, you are right. There is no need to make any rounding, ie it's not necessary to override the default gcc method for rounding floating point numbers which may vary depending on the FPU used by the system.

Test:
```C
/* gcc -Wfloat-conversion test2.c -o test2 */
#include <stdio.h>
#include <limits.h>

int main (int argc, char **argv)
{
  unsigned short i=0, temp;
  double d;

  do {
    d = (double)i/USHRT_MAX;
    temp = (unsigned short)(d * USHRT_MAX);
    if (i != temp)
      printf ("%hu %hu%s\n", i, temp, i-temp != 0 ? " x" : "");
    i++;
  } while (i!=0);

  return 0;
}
```

rbuj: I got the same result if simple precision is used instead of double precision which usually is more expensive.
```C
/* gcc -Wfloat-conversion test2.c -o test2 */
#include <stdio.h>
#include <limits.h>

int main (int argc, char **argv)
{
  unsigned short i=0, temp;
  float d;

  do {
    d = (float)i/USHRT_MAX;
    temp = (unsigned short)(d * USHRT_MAX);
    if (i != temp)
      printf ("%hu %hu%s\n", i, temp, i-temp != 0 ? " x" : "");
    i++;
  } while (i!=0);

  return 0;
}
```

PR #640, opened 2020-08-10
compositor-xrender: Unscaling from gdouble rgba values to XRenderColor 

rbuj:


vkareh (approved):
build fine for me

raveit65 (approved):
Builds and runs fine, and fixes the warnings.

PR #663, opened 2021-01-30
window: Do not render shadows on left/right tiled windows

flexiondotorg:
This pull request prevents shadows being rendered for left and right side titled windows. This behaviour is consistent with maximised windows, which also do not render shadows.

The rationale for this change is so that when two windows are titled along side each other, it prevents central shadows bleeding into the touching points of the windows.

metacity-theme-x.xml has provision to style left/right titled windows. This patch makes it possible to create window themes that present clean side-by-side tiled windows, as shown below:

no-side-tiled-shadows

### DIFF
--- a/src/compositor/compositor-xrender.c
+++ b/src/compositor/compositor-xrender.c
@@ -1022,6 +1022,19 @@ window_has_shadow (MetaCompWindow *cw)
           return FALSE;
         }
 
+      /* Do not add shadows for left/right tiled windows */
+      if (meta_window_is_tiled_left (cw->window))
+        {
+          meta_verbose ("Window has no shadow because it is tiled left\n");
+          return FALSE;
+        }
+
+      if (meta_window_is_tiled_right (cw->window))
+        {
+          meta_verbose ("Window has no shadow because it is tiled right\n");
+          return FALSE;
+        }
+
       if (meta_window_get_frame (cw->window)) {
         meta_verbose ("Window has shadow because it has a frame\n");
         return TRUE;
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -8851,6 +8851,18 @@ meta_window_is_maximized (MetaWindow *window)
   return META_WINDOW_MAXIMIZED (window);
 }
 
+gboolean
+meta_window_is_tiled_left (MetaWindow *window)
+{
+  return META_WINDOW_TILED_LEFT (window);
+}
+
+gboolean
+meta_window_is_tiled_right (MetaWindow *window)
+{
+  return META_WINDOW_TILED_RIGHT (window);
+}
+
 /**
  * meta_window_is_client_decorated:
  *
--- a/src/include/window.h
+++ b/src/include/window.h
@@ -39,5 +39,7 @@ Window meta_window_get_xwindow (MetaWindow *window);
 MetaWindow *meta_window_get_transient_for (MetaWindow *window);
 gboolean meta_window_is_maximized (MetaWindow *window);
 cairo_region_t *meta_window_get_frame_bounds (MetaWindow *window);
+gboolean meta_window_is_tiled_left (MetaWindow *window);
+gboolean meta_window_is_tiled_right (MetaWindow *window);
 
 #endif


rbuj (approved):
LGTM

Test: debug remotely macro (ssh -X) using eclipse + use the desktop theme: Menta
```
$ CFLAGS="-g -O0" ./autogen.sh --prefix=/usr --enable-debug && make && sudo make install
```
A

B

raveit65:
> The rationale for this change is so that when two windows are titled along side each other, it prevents central shadows bleeding into the touching points of the windows.

I agree with the removing the shadow on screen edges, but when only 1 window is tiled, let say on the left screen side, than it is a pain to resize the window on the side to the middle of the screen.
If 2 tiled windows are used than always the active and focused window has an invisible border.
So removing the invisible border complete isn't needed and will bring back only the resize pain.

flexiondotorg:
> I agree with the removing the shadow on screen edges, but when only 1 window is tiled

We could add that feature via a separate pull request since the logic to support that capability is absent in Marco currently.

> let say on the left screen side, than it is a pain to resize the window on the side to the middle of the screen.

I think I understand what you're describing, and that did use to be an issue. But since @vkareh added lazy border grabbing I don't have this issue. Even with window styled via `metacity-theme-1.xml` that have 0px borders, they are easy to grab and resize :slightly_smiling_face: 

> If 2 tiled windows are used than always the active and focused window has an invisible border.

This behaviour can be controlled via `metacity-theme-1.xml` where left and right tiled windows, in focused and unfocused state, can be rendered differently; including border size or absence. I have done precisely that in a new theme I've been working on recently :slightly_smiling_face:

raveit65:
> But since vkareh added lazy border grabbing

What do you mean with lazy border grabbing?

Well, personal i don't mind if we have a shadow for invisible border function or not for side-tiled windows, because i am using tab-cycle with my keyboard to resize side- and quarter-tiled windows.
But i am pretty sure we will get reports by another users who are missing the invisible border to resize the window.
At least i didn't find to time to test PR yet, because of my new personal situation and the actual review queue is huge, so i am not sure if you want to remove only the shadow by the compositor or the invisible border too.

raveit65:
> This behaviour can be controlled via metacity-theme-1.xml where left and right tiled windows, in focused and unfocused state, can be rendered differently; including border size or absence. I have done precisely that in a new theme I've been working on recently

Link please

flexiondotorg:
> Link please

https://github.com/ubuntu/yaru/tree/hirsute/mate/metacity/src

The [metacity-theme.xml.in](https://github.com/ubuntu/yaru/blob/hirsute/mate/metacity/src/metacity-theme-1.xml.in) is templated via [meson.build](https://github.com/ubuntu/yaru/blob/hirsute/mate/metacity/src/meson.build), which just provides the colour constants at the top of the file. The window button assets are in the folders `default`, `dark` and `light`.

vkareh:
@raveit65 - This still allows for invisible borders, it just removes the visible shadow from side-tiled windows.

@flexiondotorg - Currently I'm still getting the shaddow when I first tile, but as soon as the window loses focus, the shadow is gone. Might be worth investigating why the shadow is still present initially. Also, does it make sense to remove shadow from _all_ tile positions, or just left/right? Other than that, this change makes sense to me.

raveit65:
> But since vkareh added lazy border grabbing

invisible borders = lazy border grabbing
LOL, what a weird UK translation for the technical term `invisible borders`
I am not against to remove only the borders if  `invisible borders` won't be removed.

raveit65:
> The metacity-theme.xml.in is templated via meson.build, which just provides the colour constants at the top of the file. The window button assets are in the folders default, dark and light.

Those frame styles ,eg. `<frame_style name="tiled_x_x" are part of metacity theme version 3 or 2, you should rename it to metacity-theme-3 or metacity-theme-2. See our Menta themes.
I didn't find a documentation with exact information about.

At least i don't see any issues with our themes or the themes by gtk.

raveit65 (approved):
LGTM.

raveit65:
@flexiondotorg 
There is an open question by @vkareh

raveit65:
@flexiondotorg @vkareh 
Any progress with your discussion?

raveit65:
@flexiondotorg @vkareh
Any progress with your discussion?

PR #666, opened 2021-02-07
build: by default disable also the verbose mode for autoconf

rbuj:
The default value for meson build is false.
https://github.com/mate-desktop/marco/blob/05ffd59e8b5a0c125e21f487e91fb0d456b63fa3/meson_options.txt#L1-L2


raveit65 (approved):
LGTM

PR #674, opened 2021-02-24
Do not define custom exit codes and remove exit wrapper

rbuj:


raveit65:
Just a rebase with master

PR #679, opened 2021-04-14
window: Only unmaximize when the window is already maximized.

zzhiyi:
When handling a ClientMessage event that removes _NET_WM_STATE_MAXIMIZED_HORZ
or _NET_WM_STATE_MAXIMIZED_VERT from _NET_WM_STATE, marco always restores the
window rectangle to the saved rectangle. So if an application window is already
unmaximized and the application sends such a ClientMessage event, marco will
reset the window rectangle regardless. EWMH doesn't specify what must be done
when handling such events. It seems best to avoid restoring window rectangles
like other window managers in this case.

Fix a related bug: https://bugs.winehq.org/show_bug.cgi?id=50381

This revert a change introduced by 6219f8e8bcaeefb9185a3c3f5f20de4e2fa8f18f.

Signed-off-by: Zhiyi Zhang <zzhang@codeweavers.com>

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -2771,77 +2771,81 @@ meta_window_unmaximize (MetaWindow        *window,
   window->tile_mode = META_TILE_NONE;
   window->tiled = FALSE;
 
-  /* Only do something if the window isn't already maximized in the
+  /* Only do something if the window is already maximized in the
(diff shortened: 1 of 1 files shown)


vkareh (approved):
LGTM

PR #682, opened 2021-04-30
window: Track window rectangle before tiling

vkareh:
When tiling a maximized window, we should keep track of the saved
rectangle so that tiling does not reset our window size. Otherwise,
untiling the previously maximized window will end up with an unmaximized
full-size window, rather than the original window size.

To reproduce:
1. Open a window (pluma, mate-termina, etc)
2. Maximize
3. Tile right using keyboard shortcuts
4. Restore window
  a. Before this patch, the restored window size is a full size window
  b. After this patch, the restored window size is the original size before maximizing

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -2699,7 +2699,11 @@ meta_window_tile (MetaWindow *window)
     return;
 
   if(window->tile_mode == META_TILE_LEFT || window->tile_mode == META_TILE_RIGHT)
-    meta_window_maximize_internal (window, META_MAXIMIZE_VERTICAL, NULL);
+    {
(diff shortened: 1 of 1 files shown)


raveit65:
Confirmed, and it isn't really a full site window in my test. But maybe this caused by using this PR https://github.com/mate-desktop/marco/pull/678 in my test.

raveit65 (approved):
LGTM,
does what it promise.

raveit65:
No related to PR.
I think this needs to be fix for quarter-tiling too, here pressing the `alt+F5` doesn't restore the window.

raveit65:
Can this be done for quarter-tiling too, please?

PR #683, opened 2021-05-26
iconcache: Use GDesktopAppInfo to find correct application icons

vkareh:
Some files do not report their application icons correctly in the window
properties. This patch allows the marco UI to search for the
corresponding .desktop file and render the icon in the desktop info on
both the alt-tab popup and the window mini-icon.

Before:
Screenshot at 2021-05-26 14-17-59

After:
Screenshot at 2021-05-26 14-12-26

### DIFF
--- a/src/core/iconcache.c
+++ b/src/core/iconcache.c
@@ -25,6 +25,7 @@
 #include "iconcache.h"
 #include "ui.h"
 #include "errors.h"
+#include "window-private.h"
 
 #include <X11/Xatom.h>
 
@@ -690,6 +691,7 @@ scaled_from_pixdata (guchar *pixdata,
 gboolean
 meta_read_icons (MetaScreen     *screen,
                  Window          xwindow,
+                 char           *res_name,
                  MetaIconCache  *icon_cache,
                  Pixmap          wm_hints_pixmap,
                  Pixmap          wm_hints_mask,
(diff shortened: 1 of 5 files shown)


lukefromdc (approved):
Works with Olive (video editor), which does not ship an icon. Fallback icon changes to one which seems more appropriate however, old fallback icon still shows in panel and "icon-missing" shows up in mate's main menu

Olive_in_marco
Olive_in_panel

lukefromdc:
Also note that this icon is picked separately in compiz-reloaded's gtk-window-decorator, calling code in libwnck rather than libmarco, so no effect on that.

raveit65 (approved):
LGTM,
does what it promise.

balazs-endresz:
At least `libwnck` does notice changes to `_NET_WM_ICON`. E.g. creating a new file with the following and adding the path to the `Exec` command in the `.desktop` file helps fix it for the taskbar too.

```
#!/bin/bash
/usr/bin/slack &
sleep 5
cp -n /usr/share/pixmaps/slack.png ~/snap/xseticon/current/slack.png  # snaps can't access the original location
wmctrl -lx | grep slack.Slack | cut -d' ' -f1 | xargs -I{} xseticon -id {} ~/snap/xseticon/current/slack.png
```

But I guess implementing this PR in libwnck would be even better, it looks like the code is structured somewhat similarly, see `_wnck_read_icons`.

PR #684, opened 2021-05-27
window: Track window rectangle before corner tiling

vkareh:
When corner-tiling a maximized window, we should keep track of the saved
rectangle so that tiling does not reset our window size. Otherwise,
untiling the previously maximized window will end up with an unmaximized
full-size window, rather than the original window size.


raveit65 (approved):
LGTM

PR #708, opened 2021-12-12
Fix some -Wfloat-conversion warnings

rbuj:


src/core/screen.c
```diff
@@ -1239,14 +1240,14 @@ get_window_surface (MetaWindow *window)
   /* Scale surface to fit current screen */
   if (width > height)
     {
-      max_size = MIN (MAX_PREVIEW_SIZE, current->rect.width / max_columns * MAX_PREVIEW_SCREEN_FRACTION);
+      max_size = MIN (MAX_PREVIEW_SIZE, MAX_PREVIEW_SCREEN_FRACTION * ((double) current->rect.width) / ((double) max_columns));
```
raveit65, line 1243: Should it not be ?
```
- max_size = MIN (MAX_PREVIEW_SIZE, current->rect.width / max_columns * MAX_PREVIEW_SCREEN_FRACTION);
+ max_size = MIN (MAX_PREVIEW_SIZE, ((double) current->rect.width) /  MAX_PREVIEW_SCREEN_FRACTION * ((double) max_columns));
```

src/core/screen.c
```diff
@@ -1239,14 +1240,14 @@ get_window_surface (MetaWindow *window)
...
       ratio = ((double) width) / max_size;
       width = (int) max_size;
       height = (int) (((double) height) / ratio);
     }
   else
     {
-      max_size = MIN (MAX_PREVIEW_SIZE, current->rect.height / max_columns * MAX_PREVIEW_SCREEN_FRACTION);
+      max_size = MIN (MAX_PREVIEW_SIZE, MAX_PREVIEW_SCREEN_FRACTION * ((double) current->rect.height) / ((double) max_columns));
```
raveit65, line 1250: Same here?

src/core/screen.c
raveit65: Not sure if subtraction part is done before multiplication here in C.

rbuj:
```
$ cat << EOF > polish_notation.c
#include <stdio.h>

int
main (void)
{
  printf ("%d\n", 8 / 4 * 2);
  printf ("%d\n", (8 / 4) * 2);
  return 0;
}
EOF
gcc -Wall polish_notation.c -o polish_notation
./polish_notation 
```
Result:
```
4
4
```

rbuj: The operators '*' and '/' have the same preference, so they operations are made from left to right.

rbuj: where is the subtraction?

raveit65: Opps, my translating mistake. I meant division.

> The operators '*' and '/' have the same preference, so the operations are made from left to right.

Thanks for explanation.

raveit65 (approved):
LGTM

raveit65:
@mate-desktop/core-team 
Any one like to review this one?

PR #709, opened 2021-12-13
menu: avoid deprecated gtk_menu_popup

tormodvolden:
Adapted from metacity commit 37fa0d19 by Alberts Muktupāvels

Unsure about the handling of atom__GTK_SHOW_WINDOW_MENU xclient message in meta_window_client_message() in src/core/window.c - whether more gdk_event fields must be filled in, or if we could use button_press_event_new() from display.c instead. Also unsure on how to test this path.

### DIFF
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -530,20 +530,18 @@ meta_core_get_active_workspace (Screen *xscreen)
 }
 
 void
-meta_core_show_window_menu (Display *xdisplay,
-                            Window   frame_xwindow,
-                            int      root_x,
-                            int      root_y,
-                            int      button,
-                            guint32  timestamp)
+meta_core_show_window_menu (Display              *xdisplay,
+                            Window                frame_xwindow,
+                            const GdkRectangle   *rect,
+                            const GdkEventButton *event)
 {
   MetaWindow *window = get_window (xdisplay, frame_xwindow);
 
   if (meta_prefs_get_raise_on_click ())
     meta_window_raise (window);
-  meta_window_focus (window, timestamp);
+  meta_window_focus (window, event->time);
 
-  meta_window_show_menu (window, root_x, root_y, button, timestamp);
+  meta_window_show_menu (window, rect, (GdkEvent *) event);
 }
 
 void
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -530,18 +530,18 @@ meta_core_get_active_workspace (Screen *xscreen)
 }
 
 void
-meta_core_show_window_menu (Display              *xdisplay,
-                            Window                frame_xwindow,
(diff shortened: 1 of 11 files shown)


lukefromdc (commented):
This builds fine and mostly runs as expected in a quick test except for one thing, which seems to be a problem ONLY on hidpi displays or otherwise with window-scaling-factor=2 : menus shown on right-clicking the title bar (if that option has been selected) or clicking the title bar menu icon usually get rendered out of position.
Marco_PR_Menu_Placement_Error_12-14-2021
If I set window-scaling-factor=1 these menus come up where I would expect them to, looking the same as with master.

lukefromdc:
If I click around on the titlebar with the right-click mouse button, it appears that this menu is coming up exactly twice as far right and exactly twice as far down from the top left corner as it should. This implies the window scaling factor may be getting applied twice in calculating the menu's position. If it is 1 this has no effect of course

tormodvolden:
Thanks for testing! I didn't see this as I likely have scaling factor 1. We should probably add the follow-up commits ca73ad5 and af0bbde from metacity.

tormodvolden:
Hmm, [ca73ad5c](https://gitlab.gnome.org/GNOME/metacity/-/commit/ca73ad5c8bc0c38981fa9fb8efad9bb466e1f9ac) requires [2a7e7be](https://gitlab.gnome.org/GNOME/metacity/-/commit/2a7e7be6e28a25e85b31519ca88656e81705a0b4) - starting to smell rabbit hole.

lukefromdc:
Still has problems with window-scaling-factor=2 and worse than before. Now the menu position depends on the distance of the clicked on point from the top left corner of the entire screen, not just the window

tormodvolden:
A missing root/frame coordinate calculation? Is this the META_ACTION_TITLEBAR_MENU case in ui/frames.c? This part looks a bit odd now, and maybe needs scaling.

The first commit introduces get_window_scaling_factor() in display.c used for button events, and the second commit introduces another get_window_scaling_factor() in ui.c. The third commit removes the one from display.c and introduces meta_ui_get_scale() in ui.c, but this one is no longer used after the fourth commit. It is there, but not used, in current metacity as well. So if we pick something from here we could omit that. And also suggest removing it in metacity.

However, in metacity there is also a call in meta_ui_init() to "disable the automatic scale handling", and this was removed from marco in commit f6e3326a. Maybe more importantly, this commit introduced scaling in meta_window_menu_popup() which I removed in the first commit here because I thought it was no longer needed. I think the code has diverged so much from metacity that simple cherry-picking can no longer work, and it requires better understanding and a determination about which direction to take.

vkareh:
Replaced by #798

PR #713, opened 2022-01-15
Feature: Add setting to raise windows on alt+tab popup

sparkida:
I would like to propose adding a setting for raising the window during the alt+tab cycling to accommodate the difficulty of identifying which window is being chosen. When I have multiple terminals running it currently requires a lot of mental power to identify which terminal I actually want and by raising the window we add a useful visual indicator.


raveit65:
https://app.travis-ci.com/github/mate-desktop/marco/jobs/555752722#L7627
Can you please rebase against our master to get this commit https://github.com/mate-desktop/marco/commit/cb274b4b3fe7dad642fe27a58a1f4955bcb21e1d into PR?
This should fix the build issue with debian for CI.

src/core/keybindings.c
```diff
@@ -2121,6 +2121,20 @@ process_tab_grab (MetaDisplay *display,
...
 
+  gboolean raise_windows;
+  raise_windows = meta_prefs_get_alt_tab_raise_windows ();
+  if (raise_windows)
+    {
+      Window target_xwindow;
+      MetaWindow *target_window;
+
+      target_xwindow =
+        (Window) meta_ui_tab_popup_get_selected (screen->tab_popup);
```
raveit65, line 2132: New build warning. https://app.travis-ci.com/github/mate-desktop/marco/jobs/557311248#L2717
```
core/keybindings.c:2132:18: warning: cast from function call of type 'MetaTabEntryKey' (aka 'void *') to non-matching type 'Window' (aka 'unsigned long') [-Wbad-function-cast]

        (Window) meta_ui_tab_popup_get_selected (screen->tab_popup);

                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

raveit65:
Beside from the build warning i like this new behavior.

sparkida:
@raveit65 not really sure what to do at this point to handle those warnings, all the previous occurrences are and have been logging that same warning, so if we want to get rid of them I'm guessing it will require some reworking of the `MetaTabEntryKey` and the `meta_ui_tab_popup_get_selected` function. Do you see an easy solution here?...because I am not that knowledgeable with C

src/core/keybindings.c
rbuj: PR #718

raveit65:
I can confirm that https://github.com/mate-desktop/marco/pull/718 fixes the `-Wbad-function-cast` in master. So, you like to update your PR when the other one is merged?

raveit65:
Btw, i found out that `alt+esc` and `alt+f6` does basically the same as your PR without the alt-tab popup.

sparkida:
> Btw, i found out that `alt+esc` and `alt+f6` does basically the same as your PR without the alt-tab popup.

The `alt+esc` is nice that it brings the window back to the same spot in the stack rather than keeping it raised after leaving as in my PR. I guess it would be best to have that same functionality, but with the popup. What do you think?

raveit65:
> > Btw, i found out that `alt+esc` and `alt+f6` does basically the same as your PR without the alt-tab popup.
> 
> The `alt+esc` is nice that it brings the window back to the same spot in the stack rather than keeping it raised after leaving as in my PR. I guess it would be best to have that same functionality, but with the popup. What do you think?

Yeap, i agree with you, that might be the best behavior to leave the stack as it is with alt-tab popup too ,ihmo.
And you are using a gsettings key to enable it, which avoid to change the default for other users.
Personal this would be my new default, but i've learned in 10 years of developing Mate that changing a default can trigger a big shitstorm ;)
Any way, i am not sure if most users will see that change, so let the team decide if that would be the new default or not, after your PR is finished.

raveit65:
@sparkida 
Other PR is merged https://github.com/mate-desktop/marco/commit/480cc60e3b3596a7fa8e196a08974b6086217bf0

sparkida:
@vkareh @raveit65 I fixed the issues and simplified the implementation a lot, really only a couple lines of functional changes at this point

raveit65:
```
[rave@mother marco]$ git pull https://github.com/sparkida/marco.git feat/alt-tab-raise-windows
From https://github.com/sparkida/marco
 * branch              feat/alt-tab-raise-windows -> FETCH_HEAD
fatal: Not possible to fast-forward, aborting.
```
Can you please rebase with master?

raveit65 (approved):
I seems to work now.
Needs to be tested by another reviewer....

zhuyaliang (approved):
Great change, no problems found after many tests

zhuyaliang:
@vkareh Whether code can be merged ？

lukefromdc (approved):
This is best tested in combination with fast-alt-tab, so that the window in question is not highlighted by the popup itself in such a way as to obscure which window is on top. That said, it works either way, the alt-tab selected window is raised if and only if alt-tab-raise-windows is set to TRUE in dconf-editor.  

Adding this option to mate-control-center after this is merged would be nice too, as not all end users will be familiar with dconf-editor

PR #716, opened 2022-01-26
build: fix meson build

rbuj:
```
src/meson.build:214:7: ERROR: Function does not take positional arguments.
```


raveit65 (approved):
LGTM,
thx for quick reaction.

PR #717, opened 2022-01-28
Remove TRUE, FALSE from if statement

rbuj:


cwendling (commented):
Not tested, but looks good

raveit65 (approved):
LGTM,
it's tested several weeks.

lukefromdc (approved):
Coding looks good, and this builds and runs fine in a quick test. We have another person reporting good results in several weeks testing so will merge this

PR #723, opened 2022-03-28
marco: Add cmdline option --no-keybindings for disabling all key bind…

sunweaver:
…ings directly on WM startup.

 The marco WM is used in Arctica Greeter [1]. There recently
 has a security issue been detected where people could open
 applications via marco keybindings inside the greeter (display
 manager) session.
 .
 A work-around could be evoking marco-message after marco startup
 and disable all keybindings then. However, a more preferred fix
 is provided by this patch: start-up marco with keybindings disabled
 from the beginning.


zhuyaliang (approved):

vkareh (approved):
LGTM

PR #729, opened 2022-07-25
configure.ac: The macro 'AC_HELP_STRING' is obsolete

rbuj:


raveit65:
It's only a warning, not a build failure. We are support distros with older autoconf versions.
Please do not blindly remove features without checking requirements for all distros.

raveit65:
In el8 based distros, eg rock 8 we use
```
[rave@rock8 ~]$ rpm -qa autoconf
autoconf-2.69-29.el8.noarch
```
So we need the the help string for several years.
Please read https://github.com/curl/curl/issues/6647
and their suggestions.
```
AC_HELP_STRING is now [deprecated](https://www.gnu.org/software/autoconf//manual/autoconf-2.69/html_node/Obsolete-Macros.html#Obsolete-Macros) but can be replaced with [AS_HELP_STRING](https://www.gnu.org/software/autoconf//manual/autoconf-2.69/html_node/Pretty-Help-Strings.html#AS%5fHELP%5fSTRING) - which seems to indicate that this dangerous-looking warning is completely uncalled for.

I guess we should provide a private AS_HELP_STRING macro for users of autoconf <= 2.69 and then do the replace.
````

raveit65 (changes requested):
I guess we should provide a private AS_HELP_STRING macro for users of autoconf <= 2.69 and then do the replace.

rbuj:
`AC_HELP_STRING` was tagged as obsolete on [AUTOCONF-2.57a](https://github.com/autotools-mirror/autoconf/releases/tag/AUTOCONF-2.57a) in the commit https://github.com/autotools-mirror/autoconf/commit/5958ce17565d82cf54e5936e6b4e3a16169a1c9e

raveit65 (approved):

raveit65:
After i had to rebase caja for rhel8 because of a non needed glib2 change from you during development of caja-1.23.0 i am carefully.
Honestly this war a hard and time consuming word to get caja-1.26 working for rhel8 with glib2-2.56.4-158.el8.x86_64.

rbuj:
https://github.com/mate-desktop/caja/commit/8f568265bf65f4c7bdc19738f91b4e98ed826394

lukefromdc:
AS_HELP_STRING has been around back to 2008 at LEAST, as that's when Autoconf 2.62 came out. Thus no distro that will ever ship MATE 1.26 should get a build failure from this change UNLESS a very long term support distro is updating DE's on 14 year old core libraries
http://ftp.vim.org/ftp/gnu/autoconf/
. autoconf-2.62-2.63.xdelta.sig 10-Sep-2008

lukefromdc:
Note that CentOS 8 uses
autoconf-2.69-27.el8.noarch.rpm which is much newer

autoconf-2.69-11.el7.noarch.rpm is available (or was) for still-older CentOS 7
Not sure how closely these track RHEL

lukefromdc:
https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/External-Software.html
indicates AS_HELP_STRING is supported on autoconf 2.69 as mentioned in 
https://github.com/mate-desktop/marco/pull/729#issuecomment-1194459838

As of now, this commit sets the minimum autoconf version at 2.68, which came out in 2010, 13 years ago. RHEL 8 came out in 2019, nine years later

This change will not reach any distro until marco 1.28 comes out save any rolling release using 1.27, which also would have new versions of everything else, autoconf included.

Note that this change has been made in many other packages at this point. Should we merge this one too, revert all the others, or something else? Will any 13 year old LTS be using MATE 1.28?

PR #730, opened 2022-07-30
ui: avoid 'g_object_unref (NULL)'

rbuj:


mbkma (approved):
LGTM
`gtk_icon_theme_lookup_by_gicon_for_scale` may return `NULL`:
https://developer-old.gnome.org/gtk3/unstable/GtkIconTheme.html#gtk-icon-theme-lookup-by-gicon-for-scale

Issue #301, opened 2016-12-12
Firejail/Flatpak applications display "as superuser" on window title


slazur:
Applications launched under Firejail and Flatpak include "(as superuser)" in their window's title even though they're not actually being ran as root.

More information from Firejail's issues tracker: [(as superuser) in title bar](https://github.com/netblue30/firejail/issues/258).

Debian stretch, marco 1.16.0, firejail 0.9.44.2, flatpak 0.6.14.

monsta:
It's up to them to solve this, MATE apps know nothing about Firejail or Flatpak.

smitsohu:
For the record: This is probably a result of PID namespaces, and the effect should be observable with many sandboxing tools (firejail, bubblewrap, flatpack, ...)

The PID from inside the sandbox, often a single digit number, is used to set _NET_WM_PID, and then probably the window manager checks outside the sandbox if this _NET_WM_PID belongs to the superuser (which it usually does).

davidedmundson:
(I got linked to here from a KDE bug report with a similar problem)

That's not a terrible idea, but blindly switching will break just as many places as it fixes.
Like smplayer which have mplayer and the GUI as separate processes, but deliberately have a NET_WM_PID which is "faked"

halfline:
shouldnt matter unless one of the separate processes runs as root right?

SivaMachina:
Could this be related to why some applications don't seem to quite integrate with my DE, and the ones that don't seem to integrate well seem to report as being run as "superuser".

SivaMachina:
I wonder if this issue is related to why some of the flatpaks don't fit in with the DE I am using. Kinda like if I were actually running them as root.

LorenzoAncora:
The bug is still present on `Marco 1.20.3-1` [Debian GNU/Linux 10 (buster)].
End users don't care which team fixes it, @monsta.

damianatorrpm:
I have tried LOCAL_CLIENT_PID
`https://stackoverflow.com/questions/37283179/python-pid-to-x11-window-id-using-xresqueryclientids/37309217#37309217`

This does not work either for sandboxed applications.

hamishmb:
Still a problem for me on Mint 19.3 with Flatpak apps.

AsciiWolf:
Same issue on Linux Mint 20 Mate with Flatpak applications.

stavultras:
[...]

PR #738, opened 2022-10-24
Remove "(as superuser)" text

CuBeRJAN:
Unfortunately using _NET_WM_PID isn't reliable on sandboxed applications, leading to a bug with Flatpak and Firejail as seen here:
https://github.com/mate-desktop/marco/issues/301.
This commit disables the functionality completely, fixing the sandbox issue.


CuBeRJAN:
Just a note: another alternative may be using the XResQueryClientIds X extension instead of _NET_WM_PID to get the window owner, which should be more reliable.

PR #741, opened 2022-10-24
replace _NET_WM_PID with XResGetClientPid

CuBeRJAN:
This patch adresses this issue
https://github.com/mate-desktop/marco/issues/301
tested and works well with flapak.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -164,7 +164,7 @@ if test x$have_xcomposite = xyes; then
 fi
 
 if test x$have_xcomposite = xyes; then
-  MARCO_PC_MODULES="$MARCO_PC_MODULES xcomposite >= $XCOMPOSITE_VERSION xfixes xrender xdamage"
+  MARCO_PC_MODULES="$MARCO_PC_MODULES xcomposite >= $XCOMPOSITE_VERSION xfixes xrender xdamage xres"
   AC_DEFINE(HAVE_COMPOSITE_EXTENSIONS, 1, [Building with compositing manager support])
   echo "Building with compositing manager"
 
--- a/.build.yml
+++ b/.build.yml
@@ -14,6 +14,7 @@ requires:
     - libcanberra
     - libgtop
     - libxpresent
+    - libxres
     - make
     - mate-common
     - mate-desktop
@@ -51,6 +52,7 @@ requires:
     - libxpresent-dev
     - libxrandr-dev
     - libxrender-dev
+    - libxres-dev
     - libxt-dev
     - make
     - mate-common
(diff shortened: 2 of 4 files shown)


configure.ac
```diff
@@ -164,7 +164,7 @@ if test x$have_xcomposite = xyes; then
 fi
 
 if test x$have_xcomposite = xyes; then
-  MARCO_PC_MODULES="$MARCO_PC_MODULES xcomposite >= $XCOMPOSITE_VERSION xfixes xrender xdamage"
+  MARCO_PC_MODULES="$MARCO_PC_MODULES xcomposite >= $XCOMPOSITE_VERSION xfixes xrender xdamage xres"
```
raveit65, line 167: Ok, but CI failed as you can see.
https://github.com/mate-desktop/marco/pull/741/checks?check_run_id=9070958435
https://app.travis-ci.com/github/mate-desktop/marco/jobs/586481829#L928
```
Package 'xres', required by 'virtual:world', not found
```
You need to add `libXres-devel` (example fedora) to https://github.com/mate-desktop/marco/blob/master/.build.yml#L62
Same for debian, ubuntu, archlinux.
For debian and ubuntu it's called libxres-dev.
Add `libxres` for archlinux, same like fedora only without -devel

raveit65: and you need to update meson.build https://github.com/mate-desktop/marco/blob/master/meson.build

CuBeRJAN: alright, I've updated the files and builds were successful now

muktupavels:
Did you copy that code from `metacity`?

CuBeRJAN:
> Did you copy that code from `metacity`?

yes

configure.ac
raveit65: thx

raveit65:
> Did you copy that code from `metacity`?

Something wrong with the code?

CuBeRJAN:
The "(as superuser)" is still there

Here is before the commit:
Snímek obrazovky z 2022-11-06 20-26-11
You can see OBS and Spotify both show (as superuser)

Now after the commit
Snímek obrazovky z 2022-11-06 20-26-02
Caja still shows "(as superuser)", while Spotify and OBS no longer do.

raveit65:
@CuBeRJAN 
see https://github.com/mate-desktop/marco/issues/746
You PR breaks connecting a machine with marco-1.26.1 via x2go.

krotow:
Your commit either got lost in translation or simply didn't landed in Ubuntu/Mint repos. "as superuser" is still present in Linux Mint 21.3 MATE. At least 3 fresh system installs with first time online updates. All Flatpak apps that are launched from user account, has "as superuser" in title.

Issue #647, opened 2020-09-14
mate/marco window focus selection behavior should resemble a stack


garberw:
#### Expected behaviour
when closing a window A, focus should return (pop off stack) to window that had the focus just before window A got the focus.
in other words, there should be a  stack of windows successively having the focus.
this is what normally happens in other operating systems or window managers (compiz works as expected).

#### Actual behaviour
when window A is closed behavior is unpredictable.  Often focus returns to desktop.  Behavior varies with programs and
associated windows having the focus.

#### Steps to reproduce the behaviour
open window B, open window A, close window A, see whether window B still regains focus.

#### MATE general version
garberw@electron> rpm -qa | grep marco
marco-1.24.1-1.fc32.x86_64
marco-libs-1.24.1-1.fc32.x86_64
garberw@electron> 
#### Linux Distribution
fedora 32

garberw:
https://nest.parrot.sh/packages/debian/marco/blob/debian/1.22.3-1/doc/how-to-get-focus-right.txt

this commented source code explains why I am getting this problem, so this bug report is more like a feature request.
in other words, they should return to the way gnome 2 originally worked regarding window focus.

jan-kiszka:
I see this issue with marco-1.24.1-lp152.1.1.x86_64 / libmarco-private2-1.24.1-lp152.1.1.x86_64 on OpenSUSE 15.2 as well.

I'm not sure if that is the same problem, but I can also see new windows with focus not getting into the foreground. Reproduce with Okular e.g.: open another file from an existing instance, and the focus returns to the first window, rather than the newly opened one.

All that looks to me like regressions compared to 1.20.1 where I was before (OpenSUSE 15.1) and such problems did not occur.

garberw:
https://github.com/jothan/mate-window-manager/blob/master/doc/how-to-get-focus-right.txt

this web page explains that this is an (unpopular) "feature" and not a bug.
in gnome2 when you are 
(1) in a window and you 
(2) launch a second windowed program,

after the second windowed program exits you expect 

(3) the focus to return to the first window.

Instead this update on the github website above seems to make
focus mainly dependent on where the pointer (mouse) is.
It is therefore not a bug, but makes it awkward since the
desktop is usually under the pointer, so you have to press
"alt - tab" to switch to the previous window.
Since pressing "alt-tab" over and over and over and over
is already a big nuisance for something like programming or
data entry, this mouse focus choice is not going to be a popular "feature"
[...]

Issue #251, opened 2016-01-29
strange window switching foreground/background behavior [$50]


jmssil:
I've been experiencing difficulties with marco on Linux Mint 17.3. These problems did not happen after a recent fresh install, but soon afterwards.

It's hard to explain, but basically when I click to switch windows in the task bar (window list) the selected windows is not brought to the front, it stays in the back. So, it looks like I'm clicking on the correct window, but I'm clicking in one in the background.

It looks like the foreground/background + window switching logic is wrong.

Associated with this, there is a strange shadow between the maximized windows and the task bar. I'm attaching a screenshot.

![screenshot at 2016-01-29 13 43 39](https://cloud.githubusercontent.com/assets/7854469/12677072/640d1592-c68e-11e5-8880-c0ec9e7800a4.png)
##

jmssil:
I found this happens only after running HexChat in a session, even if I close it afterwards.

evaldas325:
Hi, this is biggest annoyance for me since upgrade to 1.12.  I reproduce this bug constantly on Debian/testing desktop with dual monitor setup.  Also I installed Linux Mint 17.3 on laptop and started to reproduce this problem as well. I do not use HexChat.  I have 4 workspaces with lots of windows open.  Have no idea what program/window causes this but this affects all windows opened on current workspace.

flexiondotorg:
I have experienced this on Ubuntu MATE 15.10 with MATE 1.12 installed. But, I'm using Compiz, so I don't think Marco is to blame.

evaldas325:
Restarting Marco helps me.  If I run into this weird window behavior, I execute " marco --replace" in terminal and the problem is gone.

evaldas325:
In my previous message I said that executing "marco --replace" fixes the problem temporary until windows start switching to foreground/background again unpredictably.
However I've tried "marco --replace --no-composite" and it seems turning off compositioning solves the problem permanently.

jmssil:
> However I've tried "marco --replace --no-composite" and it seems turning off compositioning solves the problem permanently.

I seem to confirm this.

jlacroix82:
I can also confirm this bug on Ubuntu MATE 16.04 with Marco. If it helps, the machine I encounter this on has a Haswell i7 CPU, 16GB RAM, Geforce GTX 760 video card. I have two other machines, both with Intel graphics, and I've never seen this issue on those machines.

pasikarkkainen:
[...]

PR #742, opened 2022-10-28
Fix window focus stack

balazs-endresz:
This reverts https://github.com/mate-desktop/marco/commit/f96255beb (except for some unnecessary whitespace changes) to fix https://github.com/mate-desktop/marco/issues/647

I've been using this for a few days and had no issues at all. It works fine with multiple workspaces too.

### DIFF
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -2223,7 +2223,7 @@ static gboolean event_callback(XEvent* event, gpointer data)
                       meta_event_detail_to_string (event->xfocus.detail));
 
           if (event->type == FocusIn &&
-              event->xfocus.detail == NotifyDetailNone)
+              (event->xfocus.detail == NotifyDetailNone || event->xfocus.detail == NotifyPointerRoot))
             {
               meta_topic (META_DEBUG_FOCUS,
                           "Focus got set to None, probably due to "


raveit65:
I am in worry ,that we run again in [strange window switching foreground/background behavior](https://github.com/mate-desktop/marco/issues/251) with your PR.
On the other side it wasn't really clear which commit from metacity did really fix the `strange window switching foreground/background behavior` issue.
I will start to reproduce https://github.com/mate-desktop/marco/issues/647 and test your revert.

raveit65:
Well, i can reproduce https://github.com/mate-desktop/marco/issues/647
```
Expected behaviour

when closing a window A, focus should return (pop off stack) to window that had the focus just before window A got the focus.
in other words, there should be a stack of windows successively having the focus.
this is what normally happens in other operating systems or window managers (compiz works as expected).
Actual behaviour

when window A is closed behavior is unpredictable. Often focus returns to desktop. Behavior varies with programs and
associated windows having the focus.
Steps to reproduce the behaviour

open window B, open window A, close window A, see whether window B still regains focus.
```
But your PR doesn't fix the issue on my box.
Tested with 2 mate-terminal windows.
Maybe you have a better reproducer...

raveit65:
Hmm,
now i can't reproduce the issue any more with latest fedora 36 build.
I' am giving up. Please provide a clear reproducer and share your gsetinngs keys of marco.

balazs-endresz:
I've tried a new vm with Ubuntu MATE 22.04. I just noticed that I don't get this issue with any pre-installed programs but it does happen when opening  for example `gitk` from the command line. I've also noticed that alt-tab gets stuck when closing gitk. But the main issue I was trying to solve was that after closing gitk the terminal window still appeared as inactive even though I could type in it. I've tried to reproduce the various issues mentioned in the ticket here and metacity but it all seems to work with this PR. I'll give fedora a try soon too. 

[1. mate 22.04 stock - previous window active only after click.webm](https://user-images.githubusercontent.com/1246848/198857103-6c33de8a-22d5-4be2-9302-e545a5d37db2.webm)
[2. mate 22.04 stock - stuck alt tab after closing window.webm](https://user-images.githubusercontent.com/1246848/198857105-fb067ca3-99a8-4318-9b8d-3fea84b15e12.webm)
[3. mate 22.04 patched.webm](https://user-images.githubusercontent.com/1246848/198857106-a360ab29-e835-4711-bb2a-17b6fa4be9cf.webm)

balazs-endresz:
One more thing to watch out for is that mate-netbook/maximus breaks the window focus/active states in a similar way (but when opening or unmaximising windows), so make sure that's not running. I'd like to look into that too at some point. I think mate-netbook used to be installed by default, so I wonder if that could have been actually the root cause of the focus issues back then.

balazs-endresz:
On a fresh fedora 36 vm I get pretty much the same issue, with one minor difference though is that I can't type in the un-focused window, unlike on ubuntu. But this PR fixes it the same way on fedora as well.

[4. fedora 36 stock.webm](https://user-images.githubusercontent.com/1246848/198876712-d0050d62-f2b9-49b1-837a-f0b4f94a116a.webm)
[5. fedora 36 patched.webm](https://user-images.githubusercontent.com/1246848/198876714-6b961685-bde7-47ec-83b2-9c3ec40ebf0b.webm)

balazs-endresz:
I'll see if I can make this work without reverting that commit. I was just happy that I managed to make it work correctly after upgrading to 22.04 (from 18.04), but I guess there should be a better way to fix this.

balazs-endresz:
I've found another way to fix this at https://github.com/mate-desktop/marco/blob/v1.26.0/src/core/display.c#L2223 by matching not just `NotifyDetailNone` but `NotifyPointerRoot` as well. 

When I open from the terminal and then close an app e.g. `mate-tweak` then `event->xfocus.detail` is equal to `NotifyDetailNone` but if I do the same with `gitk` it's `NotifyPointerRoot`. 

Would changing that make more sense? So that `NotifyPointerRoot` would also trigger `meta_workspace_focus_default_window` (instead of reverting https://github.com/mate-desktop/marco/commit/f96255beb).

balazs-endresz:
I've updated the PR with the `NotifyPointerRoot` change instead. I've noticed that `NotifyPointerRoot` is always mentioned together with `NotifyDetailNone` in the X docs: https://tronche.com/gui/x/xlib/events/input-focus/normal-and-grabbed.html, so this does seems like the right thing to do.

(Btw, very rarely, maybe one out of 50 tries the focus is still not regained. After adding some logging it appears that `event_callback` with the `FocusOut` event is called only twice when that happens. Normally it's called many more times for some reason. This occasional focus issue was also mentioned in another ticket but I can't find that now.)

raveit65:
@mate-desktop/core-team 
Can someone else from the team review this PR, please.
I am complete busy in life.

lukefromdc:
I normally use compiz so https://github.com/mate-desktop/marco/issues/647 is not something I would notice in quick marco tests if it is intermittant. I was unable to duplicate the issue with two Pluma windows open,moving focus from one to the other. On closing the pluma window I had just moved focus to, focus returned to the previous pluma window.

For this test I used marco with compositng ON, and never at any time have I turned on focus follows mouse

lukefromdc:
Also I cannot test whether this fixes https://github.com/mate-desktop/marco/issues/301 as I have neither snappy nor flatpack installed

raveit65:
Ok, with you gitk example the foreground problem is reproducible for me now. But it isn't the focus which is broken. The problem is that the window has focus but it isn't in the foreground.
After closing gtkt the terminal window isn't in  foreground but you can type in the terminal window. That means the focus is working.  With your PR the terminal window has focus and it is in the foreground.
If you agree with that you should rename the commit. focus --> foreground

balazs-endresz:
Yep, that makes sense. I've just reworded the commit message.

muktupavels:
Can you reproduce that problem with `metacity`, `mutter`, `compiz`? None of them handles `NotifyPointerRoot`...

balazs-endresz:
`metacity` works correctly without this change because I can't reproduce an event with `NotifyDetailNone`, so `meta_workspace_focus_default_window` is never called. I get the following three events with any app I try:

```
NotifyPointerRoot	6
NotifyPointer		5
NotifyNonlinearVirtual	4
```

When I apply this patch to `metacity` it still works correctly at least.

raveit65:
I noticed another fix by this commit.
In vlc-player full screen mode the control-bar is fading in/out correctly.

lukefromdc:
Is this ready to go or is more work/review needed?

muktupavels:
That code was added to handle specific case. I doubt that your problem is caused by X protocol, so debug message now likely will be lies in some cases. Adding `NotifyPointerRoot` is not fix, it just workaround / hides real problem...

lukefromdc:
OK, I will wait on this for others to weigh in or more work to be done

raveit65 (approved):
LGTM,
Commit runs fine since several days and fixes several issues.
Ready to go for me.

lukefromdc:
A .deb package could be broken at any time by a change in an underlying library which would require recompilation. We would have to maintain that, and I can tell you from personal experience with my own custom built .debs that that happens far more often than underlying code changes that force source code changes. On Debian Unstable this can happen at any time, and there are also testing and stable to deal with, plus Ubuntu, plus Mint, etc

PR #743, opened 2022-11-03
Focus on newly opened window that was brought into foreground

### LINKED ISSUE #278: Firefox "Open Containing Folder" doesn't not open caja in foreground
Hi,
   
       Test on linux mint mate 17
        marco 1.10.2
        caja 1.10.3
  
       I noticed that after downloading a file in a maximised Firefox windows opened via an icon, if one wants to use the "Open Containing Folder" caja is open in the background. The corresponding box name in mate panel is bold. (that's not a nice behaviour ;)). Now if one keeps Firefox open, get caja in the foreground and close it, cliquing again on "Open Containing Folder" opens caja in the foreground.

      If one does the same thing but opening Firefox in a terminal or in a minimized window caja opens in the foreground (that's the user friendly behaviour).

     I think that the problem comes from the window manager but i could be wrong. On Xubuntu 16.06 the behaviour is correct "Open Containing Folder" does what the user thinks, it opens the file manager in the foreground, On Ubuntu 16.06 the file manager is open in the background.

balazs-endresz:
I don't know if this is the proper way to do this but it does fix #278 and also for example when opening sublime text from the terminal: `subl .` It sets the focus on the newly opened window in both cases.

I can reproduce this issue in `metacity` as well.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -4725,7 +4725,10 @@ meta_window_raise (MetaWindow  *window)
    * the child windows appropriately.
    */
   if (window->screen->stack == ancestor->screen->stack)
-    meta_stack_raise (window->screen->stack, ancestor);
+    {
+      meta_stack_raise (window->screen->stack, ancestor);
+      meta_window_focus (window, meta_display_get_current_time_roundtrip (window->display));
+    }
   else
     {
       meta_warning (


balazs-endresz:
I've just noticed that this takes the focus when hovering over any panel, so this is definitely not the way to fix this.

PR #744, opened 2022-11-05
Always focus and bring into foreground newly opened windows

### LINKED ISSUE #278: Firefox "Open Containing Folder" doesn't not open caja in foreground

balazs-endresz:
I've added another misguided PR recently for this: https://github.com/mate-desktop/marco/pull/743 but this one attempts fix #278 in a less stupid way. It brings to front newly opened windows, such as opening the containing folder after downloading a file in firefox, or running `subl .` from the terminal. Confusingly the latter ended up in front but not focused in some cases. This seems to help with both.

This issue is reproducible with metacity as well.

I've tested this with and without mate-netbook/maximus turned on, so it should work whether the new window opens maximised or not. Also with the initial window maximised and unmaximised as well. No issues when using multiple workspaces either. All this after logging out and in again, using the modified version of marco by default.

(I suspect this might not be the best way to fix this, I have no prior knowledge of X11, marco, or C, I've just been trying to sort out some more noticeable issues after upgrading from 18.04.)

### DIFF
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -2345,6 +2345,8 @@ static gboolean event_callback(XEvent* event, gpointer data)
         {
           window = meta_window_new (display, event->xmaprequest.window,
                                     FALSE);
+          meta_window_focus (window, meta_display_get_current_time_roundtrip (window->display));
+          meta_window_raise (window);
         }
       /* if frame was receiver it's some malicious send event or something */
       else if (!frame_was_receiver && window)


muktupavels:
Focus steal prevention was not invented to constantly annoy users...

balazs-endresz:
Ah, I didn't know about that but it did lead me here: https://github.com/mate-desktop/marco/blob/master/src/core/window.c#L1912-L1920
So the focus is prevented there with the message: `window 0x4608e22 (untitled () focus prevented by other activity; 1517930 < 1529953`. If I change the return value to FALSE then it all works fine.

I've also tried this change before: https://github.com/mate-desktop/marco/issues/278#issuecomment-399861132 - it has to do with the timestamps too but didn't help.

So I guess the immediate cause of this must be that the timestamp is not set incorrectly somewhere.

In fact the comment in `intervening_user_event_occurred` refers to a metacity issue which describes the same problem, which was supposed to be fixed: https://bugzilla.gnome.org/show_bug.cgi?id=573922

lukefromdc:
Closed in favor of
https://github.com/mate-desktop/marco/pull/742

lukefromdc:
This is for a different issue than https://github.com/mate-desktop/marco/pull/742

balazs-endresz:
And the reason why the newly opened window timestamp is older is because both `subl` and `caja` have another process running already (apparently even if there's no caja window open there still a caja process running). This is mentioned on the ticket as well: https://github.com/mate-desktop/marco/issues/278#issuecomment-401126774 It works fine if I kill those processes first.

balazs-endresz:
`reload_net_wm_user_time` is triggered when opening the window, and it's different each time, but it always an older timestamp.

balazs-endresz:
Reverting https://github.com/mate-desktop/marco/commit/49d830417d9ef75241eb63d6d08a51603cfda18f (mentioned on the ticket) doesn't actually fix the issue, but it does make it work if I open a new terminal, wait around 15 seconds, and only then open `subl .`. That's because after that revert it doesn't update the user time for the terminal window as I type, so when `reload_net_wm_user_time` is called for `subl` it still gets an older incorrect value, but because the terminal wasn't updated for a while `subl` will have a more recent timestamp, and it's allowed to get focus correctly.

So, it seems to me that the main question is why we get an incorrect timestamp for `subl .` in `reload_net_wm_user_time` .

balazs-endresz:
I have no idea why but simply adding `+1` here: https://github.com/mate-desktop/marco/blob/v1.26.0/src/core/window-props.c#L379 seems to fix this too.

balazs-endresz:
One more thing about the incorrect timestamp in `reload_net_wm_user_time`: when I open a new window for `subl` it sets the user time exactly to the last user time of the previous `subl` window. Which of course is going to be older than than the user time of the terminal where I start `subl` from (which will prevent the focus). 

For apps that do gain focus correctly I always get `no information about window` which won't prevent focus: https://github.com/mate-desktop/marco/blob/v1.26.0/src/core/window.c#L1878-L1884 

I think I'll give up on this for now, so feel free to close this PR. Changing the following line to FALSE also works great for me at least: https://github.com/mate-desktop/marco/blob/v1.26.0/src/core/window.c#L1923

balazs-endresz:
I'll close this now but noticed a few more things worth noting:

* I had been testing this with sublime text v3 but v4 does always bring itself to focus when opening it from the terminal. So I guess this must be something each app has to do itself. Maybe marco is actually working as intended?
* vscode for example specifies focus stealing explicitly too, althoug only for macOS: https://github.com/microsoft/vscode/blob/1.73.1/src/vs/platform/launch/electron-main/launchMainService.ts#L54-L65 and it always focuses the new window on linux already. I don't understand this.
* I'm still very confused about caja because focusing works differently when opening it from firefox and from the terminal. The terminal always lets it steal the focus, even if there's an existing caja process. Perhaps that needs to be fixed in caja instead? 
* The "Do not auto-maximize new windows" option in mate-tweak changes whether the new window is brought into the foreground but it's not focused in either case. A new window that's brought into the foreground should probably never be unfocused so this definitely seems like a bug. Maybe it's also different if I uninstall mate-netbook completely.
* Ubuntu 22.04 (without MATE) behaves the same way as marco. But in addition you get a message with notify-send and the icon will have some animation too. 
* In KDE the focus stealing can be adjusted more granularly: https://docs.kde.org/trunk5/en/kwin/kcontrol/windowbehaviour/index.html It might useful to do the same for marco. Especially if there was a whitelist for older apps that should always return false here: https://github.com/mate-desktop/marco/blob/v1.26.0/src/core/window.c#L1923 (mate-netbook does this too for auto-maximising windows, see `/org/mate/maximus/exclude-class` in dconf)

PR #748, opened 2023-01-17
Replace deprecated API and remove compilation warnings

zhuyaliang:
Replace deprecated API and Remove some compilation warnings

### DIFF
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -198,7 +198,7 @@ reload_modmap (MetaDisplay *display)
             {
               if (syms[j] != 0)
                 {
-                  const char *str;
+                  const char *str G_GNUC_UNUSED;
 
                   str = XKeysymToString (syms[j]);
                   meta_topic (META_DEBUG_KEYBINDINGS,
@@ -1284,7 +1284,7 @@ meta_display_process_key_event (MetaDisplay *display,
   KeySym keysym;
   gboolean keep_grab;
   gboolean all_keys_grabbed;
-  const char *str;
+  const char *str G_GNUC_UNUSED;
   MetaScreen *screen;
 
   XAllowEvents (display->xdisplay,
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -2739,7 +2739,7 @@ meta_screen_sn_event (SnMonitorEvent *event,
     {
     case SN_MONITOR_EVENT_INITIATED:
       {
-        const char *wmclass;
+        const char *wmclass G_GNUC_UNUSED;
 
         wmclass = sn_startup_sequence_get_wmclass (sequence);
 
--- a/src/ui/menu.c
+++ b/src/ui/menu.c
@@ -40,7 +40,6 @@ typedef struct _MenuData MenuData;
 typedef enum {
 	MENU_ITEM_SEPARATOR = 0,
 	MENU_ITEM_NORMAL,
-	MENU_ITEM_IMAGE,
 	MENU_ITEM_CHECKBOX,
 	MENU_ITEM_RADIOBUTTON,
 	MENU_ITEM_WORKSPACE_LIST,
@@ -49,7 +48,6 @@ typedef enum {
 struct _MenuItem {
 	MetaMenuOp op;
 	MetaMenuItemType type;
-	const char* stock_id;
 	const gboolean checked;
 	const char* label;
 };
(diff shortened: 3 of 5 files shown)


lukefromdc (approved):
Using meson and dropping these five files into a previously built directory the only build warnings from these files were two deprecation warnings. Running it directly from the directory I built it in (rather than replacing my installed marco version I didn't run into any obvious issues, but note that I am used to compiz and might not notice a subtle issue on a short test.

src/core/keybindings.c
```diff
@@ -198,12 +198,9 @@ reload_modmap (MetaDisplay *display)
             {
               if (syms[j] != 0)
                 {
-                  const char *str;
-
-                  str = XKeysymToString (syms[j]);
                   meta_topic (META_DEBUG_KEYBINDINGS,
                               "Keysym %s bound to modifier 0x%x\n",
-                              str ? str : "none",
+                               XKeysymToString (syms[j]) ? XKeysymToString (syms[j]) : "none",
```
cwendling, line 203: why? this seems less optimized, and less safe (it probably won't happen, but imagine XKeysymToString() not giving the same result on each call -- not sure if a keymap change could do that, but the point is it's not a great practice to this unless you know the call is trivial and constant)

src/core/keybindings.c
```diff
@@ -1321,17 +1317,15 @@ meta_display_process_key_event (MetaDisplay *display,
...
   keysym = 0;
-  str = NULL;
 #endif
 
   /* was topic */
   meta_topic (META_DEBUG_KEYBINDINGS,
               "Processing key %s event, keysym: %s state: 0x%x window: %s\n",
               event->type == KeyPress ? "press" : "release",
-              str ? str : "none", event->xkey.state,
+              XKeysymToString (keysym) ? XKeysymToString (keysym) : "none", event->xkey.state,
```
cwendling, line 1328: Same question, but it's worse here: `keysym` has a fake value of `0` in the `! HAVE_KXB` code path, what would be the return value of `XKeysymToString(0)`?  wouldn't it be a tad confusing?

src/core/screen.c
```diff
@@ -2739,14 +2739,10 @@ meta_screen_sn_event (SnMonitorEvent *event,
...
       {
-        const char *wmclass;
-
-        wmclass = sn_startup_sequence_get_wmclass (sequence);
-
         meta_topic (META_DEBUG_STARTUP,
                     "Received startup initiated for %s wmclass %s\n",
                     sn_startup_sequence_get_id (sequence),
-                    wmclass ? wmclass : "(unset)");
+                    sn_startup_sequence_get_wmclass (sequence) ? sn_startup_sequence_get_wmclass (sequence) : "(unset)");
```
cwendling, line 2745: same, why?  what's the improvement here?

cwendling (commented):
I did *not* test nor review everything, but here's a few concerns I have.  If they are supposed to avoid some warnings, could you explicit *which* ones, and why that would be the best solution?

lukefromdc:
I don't know much about this part of the code, as most of what I've done with march concerned understanding the libmarco portions also used by compiz, and I have not played with keybindings or the underlying code that much simply as a consequence of my workflow. My test determined only that it worked, I will leave the "is this an upgrade or not" question to those with a better understanding of this part of the code

zhuyaliang:
@cwendling 
```
[22/190] Compiling C object src/libmarco-private.so.2.2600.0.p/core_keybindings.c.o
../src/core/keybindings.c: In function 'reload_modmap':
../src/core/keybindings.c:201:31: warning: variable 'str' set but not used [-Wunused-but-set-variable]
  201 |                   const char *str;
      |                               ^~~
../src/core/keybindings.c: In function 'meta_display_process_key_event':
../src/core/keybindings.c:1287:15: warning: variable 'str' set but not used [-Wunused-but-set-variable]
 1287 |   const char *str;
      |               ^~~
[25/190] Compiling C object src/libmarco-private.so.2.2600.0.p/core_screen.c.o
../src/core/screen.c: In function 'meta_screen_sn_event':
../src/core/screen.c:2742:21: warning: variable 'wmclass' set but not used [-Wunused-but-set-variable]
```
I can remove these changes and keep these warnings temporarily

cwendling:
@zhuyaliang ah, I get it now, it's all `meta_topic()` calls, and you're building without `WITH_VERBOSE_MODE`, right?  In this case, `meta_topic()` is a no-op, so nothing uses the variable.

For this case, we can either use your code (which I don't quite like as mentioned above, but that probably is fairly harmless in practice, yet I don't really know the cost of those functions), or the variable could be behind a `#ifdef WITH_VERBOSE_MODE`:
```diff
diff --git a/src/core/keybindings.c b/src/core/keybindings.c
index 6211a94..92f8758 100644
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -194,6 +194,7 @@ reload_modmap (MetaDisplay *display)
 
           while (j < display->keysyms_per_keycode)
             {
+#ifdef WITH_VERBOSE_MODE
               if (syms[j] != 0)
                 {
                   const char *str;
@@ -204,6 +205,7 @@ reload_modmap (MetaDisplay *display)
                               str ? str : "none",
                               (1 << ( i / modmap->max_keypermod)));
                 }
+#endif /* WITH_VERBOSE_MODE */
 
               if (syms[j] == XK_Num_Lock)
                 {
```

A third option could also be shutting up the compiler, letting it know we know about that:
```diff
diff --git a/src/core/keybindings.c b/src/core/keybindings.c
index 6211a94..067136d 100644
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -196,7 +196,7 @@ reload_modmap (MetaDisplay *display)
             {
               if (syms[j] != 0)
                 {
-                  const char *str;
+                  const char *str G_GNUC_UNUSED;
 
                   str = XKeysymToString (syms[j]);
                   meta_topic (META_DEBUG_KEYBINDINGS,
```

Either way, now that I see it's for debug messages, I'm less concerned.  This said, it seems to be the default to build *with* verbose mode enabled, so it makes sense for *that* to be the best-behaving code path.

cwendling:
> This said, it seems to be the default to build _with_ verbose mode enabled, so it makes sense for _that_ to be the best-behaving code path.

My bad, the default changed a while back in 397e31879bf79861a1a21ec4e2a75017e243d34b, so the default is *not* to have verbose mode by default.

muktupavels:
> > This said, it seems to be the default to build _with_ verbose mode enabled, so it makes sense for _that_ to be the best-behaving code path.
> 
> My bad, the default changed a while back in [397e318](https://github.com/mate-desktop/marco/commit/397e31879bf79861a1a21ec4e2a75017e243d34b), so the default is _not_ to have verbose mode by default.

So verbose logging has been repurposed as something intended for embedded devices... So what do you all do now, when you need verbose output? Do you ask users to recompile marco?

cwendling:
> > My bad, the default changed a while back in [397e318](https://github.com/mate-desktop/marco/commit/397e31879bf79861a1a21ec4e2a75017e243d34b), so the default is _not_ to have verbose mode by default.
> 
> So verbose logging has been repurposed as something intended for embedded devices... So what do you all do now, when you need verbose output? Do you ask users to recompile marco?

Clearly the help string update was unfortunate.  And well, I wouldn't know, but it indeed seems a bit misguided unless it has a sizeable performance penalty affecting real use cases, but probably @rbuj knows -- yet the #666 doesn't say, and merely lines it up to the newer meson default…

zhuyaliang:
@cwendling Update completed

src/ui/menu.c
```diff
@@ -489,8 +456,43 @@ meta_window_menu_new   (MetaFrames         *frames,
...
+        device = gdk_seat_get_pointer (seat);
+
+        event = gdk_event_new (GDK_BUTTON_PRESS);
+        event->button.window = g_object_ref (window);
+        event->button.time = timestamp;
+        event->button.x = event->button.x_root = root_x;
+        event->button.y = event->button.y_root = root_y;
+        event->button.button = button;
+        event->button.device = device;
+        gdk_event_set_device (event, device);
```
cwendling, line 485: AFAICT, all (internal) code paths leading to a `meta_window_menu_popup()` start either with a `GdkEvent` or an `XEvent`.  If now a `GdkEvent` is necessary, couldn't we just forward taht to here (actually making the API itself simpler), and provide a `gdk_event_from_x_event()` kind of function that would translate one to the other? (this probably exists somewhere in GDK itself, so that could be copied over if need be)

src/ui/menu.c
```diff
@@ -267,15 +243,6 @@ static GtkWidget* menu_item_new(MenuItem* menuitem, int workspace_id)
...
 	}
-	else if (menuitem->type == MENU_ITEM_IMAGE)
-	{
-		GtkWidget* image = gtk_image_new_from_icon_name(menuitem->stock_id, GTK_ICON_SIZE_MENU);
-
-		mi = gtk_image_menu_item_new();
-
-		gtk_image_menu_item_set_image(GTK_IMAGE_MENU_ITEM(mi), image);
-		gtk_widget_show(image);
-	}
```
cwendling, line 278: I have no real opinion here.  It removes a feature, but I don't think it has that much impact. Icons lovers might disagree :)

cwendling (commented):
My previous concerns have been dealt with.

I still am not a fan of everything (e.g. this time the fake event part, that could possibly be fixed another way), but I don't see a real issue with anything.

src/ui/menu.c
zhuyaliang: Doing so will introduce a lot of code and increase the scope of changes，The implementation of ```gdk_event_from_x_event()``` is similar to that of ```menu_popup_event```.We can use existing parameters to build event，This has little impact on the whole project

src/ui/menu.c
raveit65: Please do not remove icons from menus. This is one of the special feature of MATE desktop which is missing in gnome and most Mate users want to use until someone like to switch to gtk4.

lukefromdc:
I will go even farther: a switch to GTK4 would require re-implementing icons in menus, which as you can see we've
already done locally in a number of places. I will block their removal.

Worse, I heard a couple days ago that GTK4 has now deprecated menus other than popovers and all toolbars, 
meaning any future GTK5 will be a very limited toolkit for a traditional DC requiring a massive extensions library.

src/ui/menu.c
lukefromdc: I will not accept removal of icons from menus anywhere in MATE. Re-implementing them locally to work around GTK3 deprecating them (and GTK4 if ever used not supporting them) yes, but removing them no

zhuyaliang:
I will close this PR

Issue #749, opened 2023-01-26
Window titles in mate-panel/tasklist (libwnck) show only " (as superuser)"


insaner:
Running an application as root (in my case kwrite) appends " _(as superuser)_" to the window title (so it shows _"Untitled — Kwrite (as superuser)"_). However, when viewing the entry in mate-panel, it removes everything else and shows only " _(as superuser)_" as the title. 

Debugging and tracing calls showed that `libwnck` (which is how mate-panel handles these values) uses the `_wnck_get_icon_name` call to pull this from "`_NET_WM_VISIBLE_ICON_NAME`" (you can test by seeing the difference in the displayed title between `test-tasklist` and `test-wnck`). In the marco code this seems to be set in `/src/core/window-props.c` in `set_title_text`.

I have already written a simple patch that fixes the issue for me, and will be creating a pull request.

You can test by doing:

> $ xprop WM_CLASS _NET_WM_NAME _NET_WM_VISIBLE_NAME _NET_WM_VISIBLE_ICON_NAME

(Then click on your superuser-launched window). With the problem it shows:

> WM_CLASS(STRING) = "kwrite", "kwrite"
> _NET_WM_NAME(UTF8_STRING) = "Untitled  — KWrite"
> _NET_WM_VISIBLE_NAME(UTF8_STRING) = "Untitled  — KWrite (as superuser)"
> _NET_WM_VISIBLE_ICON_NAME(UTF8_STRING) = " (as superuser)"
> 

Which instead should show (results from my patch):

> WM_CLASS(STRING) = "kwrite", "kwrite"
> _NET_WM_NAME(UTF8_STRING) = "Untitled  — KWrite"
> _NET_WM_VISIBLE_NAME(UTF8_STRING) = "Untitled  — KWrite (as superuser)"
> _NET_WM_VISIBLE_ICON_NAME:  not found.

If `_NET_WM_VISIBLE_ICON_NAME` is empty, mate-panel shows the correct title.

lukefromdc:
I am not able to duplicate this, at least not with Caja or Pluma and MATE 1.27.0 from git master. Instead, I get the normal window name
with "as superuser" appended but cut off in the mate-panel window list button. Might be a difference in libwnck version though, mine has
some changes done locally for hidpi support and to apply my theming to the window menus

zhuyaliang:
I can't duplicate the problem，Can you elaborate on the reproduction method

insaner:
[...]

PR #750, opened 2023-01-26
Show correct window title when owned by superuser. Issue #749

insaner:
Basically, it adds a check to see if the title is empty, not just null.

### DIFF
--- a/src/core/window-props.c
+++ b/src/core/window-props.c
@@ -489,7 +489,7 @@ set_title_text (MetaWindow  *window,
 
   g_free (*target);
 
-  if (!title)
+  if (!title || g_utf8_strlen (title, 2) < 1)
     *target = g_strdup ("");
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
This file produces no build warnings at all when dropped into a rebuild. Runs fine, though since I didn't get the issue with my libwnck version and panel geometry I got identical behavior w this and with master: Pluma and Caja get the normal window name with "as supereuser" appended to it. in Caja part of the "as superuser" is cut off to fit the window list button, with Pluma the window name is too long to show "as superuser" in the window list button.

zhuyaliang (approved):
The problem was fixed

cwendling:
While this indeed fixes the missing title, this loses the " (as superuser)" suffix to e.g. KCalc window.  So it's good, but also very bad.

PR #751, opened 2023-01-31
Build with verbose mode support by default

cwendling:
Verbose mode is not enabled when running Macro by default, but can be toggled dynamically as needed, e.g. for debugging a user issue, possibly even while it happens.

For some reason Meson was introduced with this option off by default (d86c29b0e518f765e7570573c76bc033fbd17f6d), and we later forwarded to Autoconf with no other rationale than getting in line with Meson (397e31879bf79861a1a21ec4e2a75017e243d34b).

I was surprised to see this disabled by default, and so was @muktupavels, see short discussion on https://github.com/mate-desktop/marco/pull/748#issuecomment-1404918784.

If this is deliberate, you can close this, and possibly explain why it is better to have this disabled by default?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -65,9 +65,9 @@ MARCO_PC_MODULES="gtk+-3.0 >= $GTK_MIN_VERSION gio-2.0 >= $GIO_MIN_VERSION pango
 GLIB_GSETTINGS
 
 AC_ARG_ENABLE(verbose-mode,
-  AS_HELP_STRING([--enable-verbose-mode],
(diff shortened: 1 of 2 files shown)


raveit65 (approved):
LGTM

raveit65:
@cwendling 
Branch has conflicts.

cwendling:
Indeed, that was because of fb7950e0a03dc72afa2fed82bf96e0fcac53afde.  Fixed now, and rebased on top of master thus.

PR #752, opened 2023-03-03
iconcache: Add GDesktopAppInfo lookup as an icon caching strategy

vkareh:
Looking up GDesktopAppInfo from the window name we can get a much better match for the icon and load it at the appropriate scale. This results in matching icons to straneous windows and better looking icons overall.

### DIFF
--- a/src/core/iconcache.c
+++ b/src/core/iconcache.c
@@ -31,7 +31,7 @@
 
 /* The icon-reading code is also in libwnck, please sync bugfixes */
 
-static void
+static gboolean
 get_fallback_icons (MetaScreen     *screen,
                     GdkPixbuf     **iconp,
                     int             ideal_width,
@@ -45,6 +45,11 @@ get_fallback_icons (MetaScreen     *screen,
    */
   *iconp = meta_ui_get_default_window_icon (screen->ui);
   *mini_iconp = meta_ui_get_default_mini_icon (screen->ui);
+
+  if (*iconp && *mini_iconp)
+    return TRUE;
+
+  return FALSE;
 }
 
 static gboolean
@@ -524,6 +529,7 @@ meta_icon_cache_init (MetaIconCache *icon_cache)
   icon_cache->wm_hints_dirty = TRUE;
   icon_cache->kwm_win_icon_dirty = TRUE;
   icon_cache->net_wm_icon_dirty = TRUE;
+  icon_cache->g_desktop_app_icon_dirty = TRUE;
 
   icon_cache->wm_hints_dirty_forced = FALSE;
   icon_cache->kwm_win_icon_dirty_forced = FALSE;
@@ -551,6 +557,7 @@ clear_icon_cache (MetaIconCache *icon_cache,
       icon_cache->wm_hints_dirty = TRUE;
       icon_cache->kwm_win_icon_dirty = TRUE;
       icon_cache->net_wm_icon_dirty = TRUE;
+      icon_cache->g_desktop_app_icon_dirty = TRUE;
     }
 }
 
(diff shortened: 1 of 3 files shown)


lukefromdc (commented):
This builds and runs fine. Fixes blurry icons on 4K monitors with window-scaling=2 but I lost use of image thumbnails as eom icons. Not my call as to how important that is

lukefromdc:
I still got the application icons and not the document/photo thumbnail icons with atril and eom with this installed. Other than that seems to work fine

raveit65:
Just testing, looks good so far. Is there an icon cache which needs to be delete first for proper testing?
Or any other test case for this PR?

lukefromdc:
Any way to bring back file thumbnail icons in such apps as eom and atril with this?

raveit65:
For proper testing i like to know.
Which icons are blurred in eom/atril ? 
A screenshot describes more than thousand words :)

lukefromdc:
In my workflow, none of them with marco from current git master. All the window icons are clear and sharp, including thumbnails for pdf, jpg files etc. I only use a few filetypes with eom and atril however.

raveit65:
Eom and atril icons are in hicolor direktory. Utilies-terminal for mate-terminal is in mate-icon-theme directory for example. Maybe this makes  a different?
Any way, i am using mate-icon-theme and i do not really see the blured /eom/atril windows icon with hidipi and Blue-submarine marco-theme with my old eyes and glasses.

raveit65 (approved):
LGTM

raveit65:
What's the status of this.
And again which icons are blurred with atril/eom with this PR?
Sceenshot please.

raveit65:
@lukefromdc 
Ok now i know what you mean with the file thumbnails.
But for me this thumbnails are too small for a picture in eom, tested with blue-submarine theme.
I can't recognize any different to a thumbnail from another picture.
But maybe your eyes are better than mine.
Also i see the picture itself in eom.
Should be the same with atril.

raveit65:
@lukefromdc 
...really a blocker for you?

lukefromdc:
Blocker no, just not a change I would have done on my own

raveit65:
@vkareh @lukefromdc 
Is that change possible?

lukefromdc:
I simply do not know. Again, not endorsing nor blocking.

lukefromdc:
Note that metacity not only has the previews but renders them sharp, we can look there for how it is done (as this is only the icon handlng) even though the decorations as a whole are entirely different

Metacity_w_Preview

raveit65:
Ok, i will test commit for myself again if this commit improve sharpness, because this is the question...
But do you really recognize a picture with the preview?
For me with damaged eyes it looks like pixel mud in your screenshot.

raveit65:
Any way, i will do a new release without it, not important enough for me.....

lukefromdc:
I recognize them no problem, my vision is at it's best from computer screen range out to about 4 feet

lukefromdc:
No image reduced that small or simply scaled that far is SHARP unless it was either created that small or is something like an .svg (or we would not have so many icon handling issues w 4K etc), but with my vision it's not pixel mud either and is easily recognized compared to the contents of the window

raveit65:
Good, than the problem are my old eyes :-)

raveit65:
In a quick test i do not see a different in sharpness of icons, eg. pluma or others of my default applications.
The icon of caja will be changed, not bad for me..... .
In result this isn't important for me as stated before.
@vkareh 
you should start to argue for yourself

vkareh:
> @vkareh
> you should start to argue for yourself

I'm _not_ a fan of this change myself. What it does is ignore the window prop hints and relies on the program name and questionable string matching to see if there's a desktop file it can see. It's not always accurate, though, and gets rid of thumbnails.

I think this should be closed.

Issue #705, opened 2021-11-30
Manually bringing windows to focus


switzel:
I do not know about the subtleties of window focus but it seems that various window managers do things consistently and marco does them differently. Concretely:

#### Expected behaviour
```
gvim --remote-silent "some-file"
```
is supposed to bring `gvim` to focus (and does with other window managers).

#### Actual behaviour
The window does not gain focus.

#### Steps to reproduce the behaviour
Open `gvim`, change focus to another window, enter the above line.

#### MATE general version
Mate 1.24.0

#### Package version
1.24.0-1ubuntu1

#### Linux Distribution
Ubuntu 20.04.3 LTS

PR #753, opened 2023-03-09
fix switch-windows-all hotkey

dnk:
pass focus to window if it is on another workspace.

current behavior:
1. there are 2 windows on workspace A and window 2 is above window 1
2. switch to workspace B
3. use "switch-windows-all" keys to switch to window 1

actual result:
window 1 get "demands attention" mark.
focus passed to window 2.

with this changes:
focus passed to window 1. window 1 is raised above all normal windows.

there is still problem with stealing focus by "always on top" windows. it can fixed separately.

### DIFF
--- a/src/core/keybindings.c
+++ b/src/core/keybindings.c
@@ -2106,8 +2106,6 @@ process_tab_grab (MetaDisplay *display,
                       target_window->desc);
           display->mouse_mode = FALSE;
           meta_window_activate (target_window, event->xkey.time);
-          if (!target_window->on_all_workspaces)
-            meta_workspace_activate (target_window->workspace, event->xkey.time);
 
           meta_topic (META_DEBUG_KEYBINDINGS,
                       "Ending grab early so we can focus the target window\n");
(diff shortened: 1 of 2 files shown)


dnk:
possible can fix https://github.com/mate-desktop/marco/issues/705

dnk:
known problem:
scenario:
there are 2+ monitors
workspace 1: 
* application A on monitor 1
* application B (Atom text editor) on monitor 2
active workspace 2
switch to application A using "switch-windows-all" hotkey 

actual result:
application B (Atom text editor) on monitor 2 displays invalid image.

the same problem appeared by just switching workspaces.

may be there is bug in Atom text editor.

Rambo1377 (approved):
src/core/window.c

lukefromdc:
This isn't something I can easily test as its so far outside my workflow. I have never used a switch-windows-all hotkey and the entire described workflow is out of my experience.It would not be visually obvious to me if this was working right or not.

We have one approving review and the code change itself is simple. Hoping we can get a second review on this.

vkareh (changes requested):
I can confirm the issue. However, in your comments you mentioned that "always on top" windows will be addressed separately, but this commit seems to only address "always on top" windows.

In either case. The changes here do not fix the issue for me on either regular windows or "always on top" windows.

PR #758, opened 2023-07-18
Fix losing title of windows with a 3 or 4 bytes first character

### LINKED ISSUE #757: Window title disappears when the first character is Chinese
#### Expected behaviour
Marco displays a window title starting with a character >= 3 bytes long.

#### Actual behaviour
Marco does not display a window title starting with a character >= 3 bytes long.

Open https://zh.wikipedia.org with firefox. The titlebar of firefox window does not contain a title.

#### MATE general version
1.26.1

#### Package version
marco 1.26.2

#### Linux Distribution
Debian

#### Link to bugreport of your Distribution (requirement)
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1041196

It seems that the fix for #749 caused this. Most Chinese characters are 3 bytes long under UTF-8, and UTF-8 characters can be as long as 4 bytes. So "g_utf8_strlen (title, 2) < 1" will be true if the first character is Chinese.

cwendling:
Fixes #757.

### DIFF
--- a/src/core/window-props.c
+++ b/src/core/window-props.c
@@ -489,7 +489,7 @@ set_title_text (MetaWindow  *window,
 
   g_free (*target);
 
-  if (!title || g_utf8_strlen (title, 2) < 1)
+  if (!title || !*title)
     *target = g_strdup ("");
   else if (g_utf8_strlen (title, MAX_TITLE_LENGTH + 1) > MAX_TITLE_LENGTH)
     {


lukefromdc (approved):
Builds clean, and fixes the issue though with my fonts I only get the missing-character symbols. Without this the space stays entirely empty however, so this should confirm it works. Someone with the appropriate fonts installed ideally would also test this, but I doubt the characters themselves would change with this as opposed to whether they are displayed

cwendling:
@lukefromdc it shouldn't change anything, but here I get proper characters (well, I see Chinese characters, and assume it's the right ones :)) and it works

raveit65:
I have no idea what's wrong with https://zh.wikipedia.org/ on my fedora system. Chinese symbols are in the middle of the window title. Without a screenshot from reporter i would say it's debian specific :)

raveit65:
I can confirm now the issue, sadly the bad commit is in latest 1.26. release.
I will test PR, but without a clear confirmation for PR from someone who use Chinese symbols, i don't like to do a new release for one commit to avoid breaking stable 1.26 branch again.

cwendling:
@raveit65 hopefully @cloudfish will be able to test, but this is a very simple fix I am confident in.  If you confirm it fixes the Firefox example for you as well we should be good.  A maybe more important test is whether it doesn't break #749 -- but my tests with KCalc tell me it doesn't -- and the logic I see agrees.

raveit65 (approved):
Confirmed that it fixes the issue
and it doesn't break previous fix for running applications as superuser.

raveit65:
If it is urgent Debian maintainer can use it as patch for his build.
Same what i do in fedora when it is urgent.
I like to test things a while before they go in stable.
Btw. everybody from the team can do new releases. This is not only in my hand ;)

cwendling:
@mate-desktop/core-team can I cherry-pick this in 1.26?  We have the breaking change there, and it's affecting distros that might or might not know about it.

cwendling:
Cherry-picked as a938884530839152c9854dd0d8248b36c3c2e3fd

PR #761, opened 2023-10-01
replace deprecated gtk_image_menu_item

raveit65:


raveit65:
Hmm, the popup is between top and bottom monitor now.

raveit65:
Weird, when using 2 monitors but position seems Ok between 2 monitors.

lukefromdc (approved):
Builds fine, and the icons (there are only a few) in the window management menu show the same as before. Both meson and autotools tested

Issue #762, opened 2023-10-08
raise-on-click "buggy behavior"


0n-s:
#### Expected behaviour
Windows should be able to raise themselves even when `org/mate/marco/general/raise-on-click` is `false`.

#### Actual behaviour
The "buggy behavior" mentioned in the description of the GSettings key `org/mate/marco/general/raise-on-click` is reproducible.

#### Steps to reproduce the behaviour
See description of the aforementioned GSettings key. This links to the [original Metacity bug](https://bugzilla.gnome.org/show_bug.cgi?id=445447), with a reproducer in the 1st comment.

I would like to mention that Metacity got a [bugfix for this issue](https://gitlab.gnome.org/GNOME/metacity/-/commit/50358a95cf7dacdb8c0c5f9ad596305298ed6f8a) shortly before it was put to rest. This patch applies cleanly to Marco without modification & that's what I've been doing on my system to fix this bug, but it'd be nice to see it fixed for everyone since it's already been fixed in the original for over a decade now.

#### MATE general version
1.26.2

#### Package version
1.26.2

#### Linux Distribution
NixOS

raveit65:
Pull request please?

raveit65:
> See description of the aforementioned GSettings key. This links to the [original Metacity bug](https://bugzilla.gnome.org/show_bug.cgi?id=445447), with a reproducer in the 1st comment.

Please write it here....

0n-s:
> > See description of the aforementioned GSettings key. This links to the [original Metacity bug](https://bugzilla.gnome.org/show_bug.cgi?id=445447), with a reproducer in the 1st comment.
> 
> Please write it here....

Sure.

> Consider the following Tk application:
>
> ```
> #!/usr/bin/env wish
>
> button .b \
>    -text "Click to raise window 2" \
>    -command "raise .t"
> pack .b
>
> toplevel .t
> pack [label .t.l -text "window 2"]
>
> wm group .t .
> ```
> Run it.  Move the button window over the top of "window 2".  Press the button.  If  the option raise_on_click is set, it works correctly.  If the option is unset, the raise event seems to be ignored, and "window 2" remains below the button window.

raveit65:
Ok, the reproducer does what it promised.

But i can't remember that i ever run in this bug (if it really is one) with an application because `raise-on-click` is set to true on my systems.
I never changed the default setting.

0n-s:
[...]

PR #763, opened 2023-10-10
Allow applications to raise windows when raise_on_click is off.

0n-s:
This is Metacity commit 50358a95 ("Allow applications to raise windows when raise_on_click is off. Closes #445447.") applied to Marco without modification.

It also includes a change to the GSettings key's description to remove the now obsolete warning on the raise-on-click key, & replaces it with an actually useful description. This is copied from the equivalent key in gsettings-desktop-schemas.

Fixes: https://github.com/mate-desktop/marco/issues/762


lukefromdc:
I can test this for build, for not crashing, and not changing my normal work flow but I never, ever experiment with focus folllows mouse as I cannot work that way and over the years have seen reports of bugs getting back out of it.

lukefromdc (approved):
Builds and runs fine, no surprises in a quick (not extensive) test with my normal click to focus settings. Without unsetting focus-mode=click (a different gsettings preference) windows are of course still raised on click regardless of the setting of this.

Didn't get the focus issues sometimes seen on restarting the WM though, but I do not know if that's a benefit of this PR (perhaps setting not applied yet on startup?) or just not having enough windows open to invoke that issue.

raveit65 (approved):
LGTM

Issue #543, opened 2019-09-05
Workspace Switcher confused


ajfre:
#### Expected behaviour
Dialog Windows, which are not "Always on Visible Workspace" should disappear on a workspace switch.

#### Actual behaviour
The dialog window also appears after switching the workspace and the application itself appears minized in the tray of the new workspace.
After closing the dialog window, the application cannot be unminimized.
Switching back to the former workspace, the app has completely disappeared there.
Switching back once more shows the app unminimized.

#### Steps to reproduce the behaviour
1. Start synaptic in workspace 3
2. click search - the search dialog opens
3. switch to workspace 5 - the search dialog appears and synaptic is minimized
4. close the search dialog
5. try to unminimize synaptic - it dosn't unminimize
6. switch to workspace 3 - there is no synaptic any more
7. switch back to ws 5 - synaptic is unminimized

Synaptic is only an example, the issue can also be reproduced with caja's copy file progress dialog.

See the attached screencast
[Screencast 2019-09-05 01:53:45.mp4.zip](https://github.com/mate-desktop/mate-applets/files/3576920/Screencast.2019-09-05.01.53.45.mp4.zip)

#### MATE general version
Don't know where to find the version ID. I reproduced the issue on Ubuntu Mate 18.04.3, Linux Mint 19.1 and 19.2

#### Linux Distribution
Ubuntu Mate 18.04.3, Linux Mint 19.1 and 19.2

sc0w:
issue transferred to marco

I confirm the same issue with the progress dialog in engrampa

raveit65:
@sc0w @lukefromdc 
Sounds like that this is fixed by https://github.com/mate-desktop/marco/commit/bf89c4c4221f060006a2fef93c72ce9d949adcec  , or not?

lukefromdc:
A progress dialog should not be making a legacy fullscreen request, and what I've often seen is dialogs that force "Always on visible workspace" with synaptic as one of the offenders. I only see fixes for dealing with `meta_prefs_get_force_fullscreen` in https://github.com/mate-desktop/marco/commit/bf89c4c4221f060006a2fef93c72ce9d949adcec

ajfre:
This is not the only bug of mate in conjunction with workspace switcher: quite often mate loses track of the z-order of the windows after workspace switching. For example, if you have several editors on top of each other and paste text, it happens that the text is not inserted in the displayed window, but in one below it.

[...]

PR #768, opened 2024-02-09
workspace: ignore not_this_one if not in current workspace

bitness:
When the active workspace is changed, marco figures out which window should get focus by calling `focus_ancestor_or_top_window`. In some cases that call might include a window that should not get focus as `not_this_one`.

When `not_this_one` refers to a window, the function will check to see if it has a parent, and if it does, it will ignore the new workspace's `mru_list` and will focus that parent window. However, it doesn't check to see if the parent window is actually on the new workspace. If the parent isn't on the new workspace, focusing it will drag it over, including the transient window that was supposed to be ignored. This isn't the result a user would likely expect, and is made more confusing by the parent window being minimized, stuck that way until the user switches to another workspace and back.

This change makes `focus_ancestor_or_top_window` ignore the `not_this_one` window if it isn't on the new workspace's `mru_list`. Instead it will just search the new workspace's `mru_list` for a window to focus.

### DIFF
--- a/src/core/workspace.c
+++ b/src/core/workspace.c
@@ -1049,8 +1049,9 @@ focus_ancestor_or_top_window (MetaWorkspace *workspace,
     meta_topic (META_DEBUG_FOCUS,
                 "Focusing MRU window\n");
 
-  /* First, check to see if we need to focus an ancestor of a window */
-  if (not_this_one)
+  /* First, check to see if we need to focus an ancestor of a window
+   * that exists in this workspace */
+  if (not_this_one && g_list_find (workspace->mru_list, not_this_one) != NULL)
     {
       MetaWindow *ancestor;
       ancestor = NULL;


bitness:
This PR should resolve issue #543 .

lukefromdc (approved):
Simple fix and confirmed here to work with the synaptic search dialog.

Issue confirmed in initial test, note that this is a VERY old bug,  I've seen it in compiz too, and regarded it as quirky but normal behavior. Nice to finally have a fix

raveit65:
I do not use debian based distros with synaptics . I use fedora and other redhat family distros.
With your python test file the dialog is only visible on the origin workspace. When i switch the workspace i do not see the open dialog. When switching back to origin workspace the open dialog is there.
Honestly , i don't understand the issue and i have no idea to reproduce the steps-to-reproduce in report.

lukefromdc:
No need to use Synaptic. Any similar dialog should do and the caja progress dialog was reported to act this way too. Yes I have seen that myself

bitness:
We also use Red Hat-family distros. We found the issue while using the marco package in EPEL's el9 repository. I'm attaching a video here showing it happening while using that RPM on an EL9 workstation.

https://github.com/mate-desktop/marco/assets/318366/3a3b3dab-43c6-43ad-ad9c-34d24af00385

bitness:
In the video, I'm first showing that the issue does not happen if the dialog does not have focus. Then the problem reproduces once I've focused the dialog.

raveit65:
OMG, it' been a long time a go that i switch workspaces without keyboard shortcut. :-)
With using the mouse i see the issue.
I will test your fix now.

raveit65 (approved):
LGTM

PR #770, opened 2024-02-16
themes/eOS/meson.build: Install missing PNGs

bobby285271:
Follow-up of 4a2a5dc092814bfe ("eOS theme: Add single button backgrounds")

Without this fix these PNGs are installed when using autotools but not using meson.


raveit65 (approved):
LGTM,
thx

PR #772, opened 2024-02-21
tabpopup: Report the window as active to a11y when it's showing

### LINKED ISSUE #771: Marco window lacks accessible "active" state, makes Orca laggy in Alt+Tab
Steps to reproduce:

1. Open several windows
2. Put the following accessible-event listener script in a file, then launch it in a terminal:

```
#!/usr/bin/python3

import pyatspi

def listener(e):
    try:
        if e.source.getApplication().name.lower() != "marco":
            return
    except:
        return

    window = pyatspi.findAncestor(e.source, lambda x: x and x.getRole() == pyatspi.ROLE_WINDOW)
    if window:
        isActive = f"container {window} is active: {bool(window.getState().contains(pyatspi.STATE_ACTIVE))}"
    else:
        isActive = "containing window not found"
    print(f"'{e.any_data}' {isActive}")

pyatspi.Registry.registerEventListener(listener, "object:text-changed:insert")
pyatspi.Registry.start()
```

3. Alt+Tab to switch among the open windows

Expected results: As the listener prints out the current switcher item, it would state that the switcher window is active.

Actual results: As the listener prints out the current switcher item, it would state that the switcher window is not active.

Sample output:

```
$ ./marco.py 
'jd@builder:~/Desktop' container [window | ] is active: False
'marco.py (~/Desktop) - gedit' container [window | ] is active: False
'debug.out (~) - gedit' container [window | ] is active: False
'Synergy 3 RC3 - v3.0.77.2-rc3' container [window | ] is active: False
'Unsaved Document 1 - Pluma' container [window | ] is active: False
'marco.py (~/Desktop) - gedit' container [window | ] is active: False
'debug.out (~) - gedit' container [window | ] is active: False
'Synergy 3 RC3 - v3.0.77.2-rc3' container [window | ] is active: False
'Unsaved Document 1 - Pluma' container [window | ] is active: False
'jd@builder:~/Desktop' container [window | ] is active: False
'marco.py (~/Desktop) - gedit' container [window | ] is active: False
'debug.out (~) - gedit' container [window | ] is active: False
'Synergy 3 RC3 - v3.0.77.2-rc3' container [window | ] is active: False
'Unsaved Document 1 - Pluma' container [window | ] is active: False
'jd@builder:~/Desktop' container [window | ] is active: False
'marco.py (~/Desktop) - gedit' container [window | ] is active: False
'debug.out (~) - gedit' container [window | ] is active: False
'Synergy 3 RC3 - v3.0.77.2-rc3' container [window | ] is active: False
```

Impact: Orca is laggy when users press Alt+Tab. The reason why is that Orca checks (and if need be, searches for) the active window. It does this in order to ensure the correct script is being applied to the accessible event. Unfortunately, when using MATE + Marco, the active window is the switcher. But it lacks the active state. So Orca checks every single open window to find the active one. It fails because none of the other windows are active; the Marco window is. But Orca doesn't know that.

It would be great if this could be fixed in Marco. I (the Orca maintainer) am removing hacks/workarounds from Orca and don't plan to add new ones.

@cwendling: Could you please verify -- and ideally fix -- this issue? Thanks in advance!!

cwendling:
The switcher windows aren't actually "active" per GTK's meaning because they do not have actual keyboard focus, but they are controlled by the internal grabs so it's effectively the same as if they were active.

Reporting them as such helps the ATs understanding what's going on.

Fixes #771.

### DIFF
--- a/src/ui/tabpopup.c
+++ b/src/ui/tabpopup.c
@@ -408,13 +408,6 @@ meta_ui_tab_popup_new (const MetaTabEntry *entries,
 
   popup->label = gtk_label_new ("");
 
-  /* Set the accessible role of the label to a status bar so it
-   * will emit name changed events that can be used by screen
-   * readers.
-   */
-  obj = gtk_widget_get_accessible (popup->label);
-  atk_object_set_role (obj, ATK_ROLE_STATUSBAR);
-
   gtk_widget_set_margin_start (popup->label, 16);
   gtk_widget_set_margin_end (popup->label, 16);
   gtk_widget_set_margin_top (popup->label, 0);
@@ -488,6 +481,15 @@ meta_ui_tab_popup_new (const MetaTabEntry *entries,
   gtk_label_set_text (GTK_LABEL (popup->label), "");
   /* Make it so that we ellipsize if the text is too long */
   gtk_label_set_ellipsize (GTK_LABEL (popup->label), PANGO_ELLIPSIZE_END);
+
+  /* Set the accessible role of the label to a status bar so it
+   * will emit name changed events that can be used by screen
+   * readers.
+   * We do this *after* messing with the label content for computing the size
+   * not to send a change even for each entry at start.
+   */
+  obj = gtk_widget_get_accessible (popup->label);
+  atk_object_set_role (obj, ATK_ROLE_STATUSBAR);
 
   int default_window_width = 0; /* 0 == small as possible, truncate label */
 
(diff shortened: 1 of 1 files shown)


cwendling:
@joanmarie I just added an extra change to be less spammy when the popup first appears, so you should not get 1 changed even per open app listed in the popup (or per workspace in the workspace switcher). Tell me if that breaks something, but I guess it'll only be better that way :)

cwendling:
Closing in favor to #773 for CI purposes (although it looks like CI is updating here *as well*, who knew…)

PR #776, opened 2024-04-05
iconcache: Use desktop_id to lookup GDesktopAppInfo

vkareh:
This replaces #752 by using either `_BAMF_DESKTOP_FILE` or `_GTK_APPLICATION_ID` to reconstruct the path to the desktop spec file. The effect is that any application that reports either of these property to the window (like many GNOME applications do), and has a matching desktop file, will be picked up correctly and Marco will get the best possible icon for that window.

If a desktop ID cannot be found for the window, it moves on to the next matching method, so nothing changes for them.


lukefromdc (approved):
Most of the code here I don't know a lot about, but this didn't break thumbnailing in atril or in EOM, in both cases clear, sharp thumbnails when available and clear, sharp icons in general

lukefromdc:
Rebased against master to pick up Travis fix, then force-pushed

PR #747, opened 2022-12-19
Test Xres extension before using it.

joakim-tjernlund:

Issue #746, opened 2022-11-24
Code dump when accessing from X2go with latest version (1.26.1)


greenfoo:
Today we upgraded our servers from Rocky 8.5 to Rocky 8.7, which includes "marco 1.26.1"

When accessing these servers from X2Go, windows are missing the "Close", "Minimize", "Maximize", ... buttons. Running "marco" from the terminal shows a "free(): Invalid pointer" message and produces a core dump.

Downgrading "marco" to the previous version ("1.22") fixes the issue.

I'm sorry I cannot provide more details or dig deeper into the problem (these are production servers and I cannot experiment with them now that we have found a "workaround")... If I'm posting this is in case someone else with the same issue finds this information useful.

Thanks and sorry for the lack of details.

raveit65:
marco- 1.22 isn't from epel repo. I recommend disable corps repo (non fedora standard).
Maybe this is a result of conflicting repos?
Beside from that marco on rocky8 works fine using VM and connecting with spice channel.
So i guess it's more a problem with x2go.
Sorry, i don't use use x2go, and my spare time is seriously limited nowadays. I can't test software which i don't use.
And sorry again, without a stacktrace no one can help you.

lukefromdc:
Wonder if rhis is yet another xpresent issue given the timing?

raveit65:
Sounds similar to libwnck issue https://gitlab.gnome.org/GNOME/libwnck/-/issues/154

raveit65:
marco-1.24.2 with reduced glib required version runs fine with rocky.
```
From 4c44f6d991065a0a37116bd90fd727867fa08d96 Mon Sep 17 00:00:00 2001
From: raveit65 <mate@raveit.de>
Date: Fri, 24 Jun 2022 19:48:29 +0200
Subject: [PATCH] glib: build with 2.56.4 for epel8

---
 configure.ac          | 2 +-
 src/core/eventqueue.c | 7 +++++++
 2 files changed, 8 insertions(+), 1 deletion(-)

diff --git a/configure.ac b/configure.ac
index 6b946ab6..498da2e6 100644
--- a/configure.ac
+++ b/configure.ac
@@ -106,7 +106,7 @@ AC_ARG_ENABLE(shape,
 ## try definining HAVE_BACKTRACE
 AC_CHECK_HEADERS(execinfo.h, [AC_CHECK_FUNCS(backtrace)])
 
[...]

PR #786, opened 2025-01-07
Fix client pid check

### LINKED ISSUE #301: Firejail/Flatpak applications display "as superuser" on window title

cwendling:
Fixes #301.

---

Disclaimer: I didn't test this, I don't actually use flatpacks or such.  However, given a check was clearly the wrong way around, and I imported another fix from Metacity, it might just work.

If this doesn't do it, possibly importing Metacity commits dade470e13292c58822b0c180019caf0ea3862f3 c6584a38a234726c1128d044fd7cc5a0f8379c69 c5136ec907fab167ffddfe4d10c3902c6c5d8852 8430e8436cb6bac732c00de5b5a3c3d73be62871 a7d6f11b3e712751b010a8f7622b148c93765f70 7c1cc3ca1d8131499b9cf2ef50b295602ffd6112 3e6358113acf61e0c68419bec6cc68a29603a2ec might help (it's mostly refactoring in the end tho, so I'm hopeful the simpler patch here would work).

CC @lukefromdc @lambdanil @tidux @rusty-snake @caiodev (and everybody else from #301), please test if you can :)

### DIFF
--- a/src/core/window-props.c
+++ b/src/core/window-props.c
@@ -320,11 +320,10 @@ get_local_pid (MetaWindow *window)
 
   for (i = 0; i < num_ids; i++)
     {
-      if (client_ids[i].spec.mask == XRES_CLIENT_ID_PID_MASK)
-        {
-          pid = XResGetClientPid (&client_ids[i]);
-          break;
-        }
+      pid = XResGetClientPid (&client_ids[i]);
+
+      if (pid != -1)
+        break;
     }
 
   XResClientIdsDestroy (num_ids, client_ids);
--- a/src/core/window-props.c
+++ b/src/core/window-props.c
@@ -311,12 +311,12 @@ get_local_pid (MetaWindow *window)
   spec.client = window->xwindow;
   spec.mask = XRES_CLIENT_ID_PID_MASK;
 
-  if (!XResQueryClientIds (window->display->xdisplay,
+  if (XResQueryClientIds (window->display->xdisplay,
                           1,
                           &spec,
                           &num_ids,
-                          &client_ids))
-      return pid;
+                          &client_ids) != Success)
+    return pid;
 
   for (i = 0; i < num_ids; i++)
     {


raveit65:
Run as superuser is gone in titlebar with signal messenger from flathub.
Hopefully you do not run again in this bug caused by first attempted tu fix this issue.
See https://github.com/mate-desktop/marco/issues/746 and reverted commit https://github.com/mate-desktop/marco/commit/d3add65f6960dd3c95aaa2fe63cc8c563747695d

cwendling:
@raveit65 I guess the issue you mentioned under X2Go was due to the lack of verification both that XRES was available, and that the XRES call was successful or not.  Unfortunately although the XRES test implemented in #747 was good, the added check for the whether the XRES call was successful was done the wrong way around.

Either way, this shouldn't bring back the X2Go issue as both checks are kept, only the latter is fixed so things actually work when XRES *is* present.

And thanks for testing :)

lukefromdc:
I cannot test this for function (no flatpack) but can check the build easily enough

lukefromdc (commented):
While I could not test with flatpak (not installed here), I didn't get any issues with the build, nor a problem with not showing caja under sudo as "ROOT" or applying that to caja running as a normal user

cwendling:
@caiodev have you had a chance to test this yet?

mariafrancesca188 (approved):

lukefromdc:
Is this ready to go or are we still waiting on another review?

cwendling:
I think this can get in, although I would have liked people that cared most in the original issue to test… but well.
Anyway, at least 3 of us tested this successfully; and I believe it has little risk of introducing additional issues (basically it fixes a check that was the wrong way around).

caiodev:
So sorry for my delay, guys.

I tried to build it using @cwendling instructions but that's the output I got:

`/usr/bin/mate-autogen
/usr/bin/yelp-build
checking for autoreconf >= 2.53...
  testing autoreconf... found 2.71
configure.ac:44: warning: The macro `AC_HEADER_STDC' is obsolete.
configure.ac:44: You should run autoupdate.
./lib/autoconf/headers.m4:704: AC_HEADER_STDC is expanded from...
configure.ac:44: the top level
checking for pkg-config >= 0.14.0...
  testing pkg-config... found 1.8.1
checking for mate-common >= 1.1.0...
  testing mate-doc-common... found 1.26.0
Checking for required M4 macros...
Checking for forbidden M4 macros...
**Warning**: I am going to run `configure' with no arguments.
If you wish to pass any to it, please specify them on the
`./autogen.sh' command line.

Processing ./configure.ac
configure.ac:44: warning: The macro `AC_HEADER_STDC' is obsolete.
configure.ac:44: You should run autoupdate.
./lib/autoconf/headers.m4:704: AC_HEADER_STDC is expanded from...
configure.ac:44: the top level
Running mate-doc-common...
Running autoreconf...
autoreconf: export WARNINGS=no-portability
autoreconf: Entering directory '.'
autoreconf: running: autopoint --force
Can't exec "autopoint": No such file or directory at /usr/share/autoconf/Autom4te/FileUtils.pm line 293.
autoreconf: error: autopoint failed with exit status: 2`

Before this attempt, I had to install mate-common and yelp-tools from official servers (Should they be from another source?) too but I couldn't build it to a working point

Do I have to give those files execution permission? Btw, I went to clone your branch @cwendling but couldn't find one branch named client-pid so I cloned this repo and there it was. Would you guys have any idea what could be wrong on my end?

Much appreciated :+1:

caiodev:
Just got a little bit further I think

I installed autopoint but now it complains about glib-2.0: Package 'glib-2.0', required by 'virtual:world', not found

Since the full log is lengthy, I put it into a text file I'll be attaching to this message :+1: 

[marco_build_log.txt](https://github.com/user-attachments/files/18576922/marco_build_log.txt)

cwendling:
@caiodev Have you run `sudo apt-get build-dep marco` (which installs all dependencies the installed version of Marco needs to be built)?  Although this only works on Debian, Ubuntu, Mint and other Apt-based distros.  Not sure what is the equivalent one others, sorry.

Anyhow, for `mate-desktop-2.0` you need a package similar to *libmate-desktop-dev* (which is Debian's name for it, but something on the lines of *libmate-desktop-devel* or so probably exists in Fedora, etc.).
**However**, the version from your distro might be too old, as current development version of Marco needs 1.27.1.  If that's the case, you then need to build that one as well (which is not harder), which you can find there: https://github.com/mate-desktop/mate-desktop/

raveit65:
Commit is already in fedora f40/41/rawhide in a few days :-)
https://bodhi.fedoraproject.org/updates/FEDORA-2025-0733c31535

caiodev:
> @caiodev Have you run `sudo apt-get build-dep marco` (which installs all dependencies the installed version of Marco needs to be built)?  Although this only works on Debian, Ubuntu, Mint and other Apt-based distros.  Not sure what is the equivalent one others, sorry (edit: for Fedora it seems there's `dnf builddep marco`).
> 
> Anyhow, for `mate-desktop-2.0` you need a package similar to *libmate-desktop-dev* (which is Debian's name for it, but something on the lines of *libmate-desktop-devel* or so probably exists in Fedora, etc.).
> **However**, the version from your distro might be too old, as current development version of Marco needs 1.27.1.  If that's the case, you then need to build that one as well (which is not harder), which you can find there: https://github.com/mate-desktop/mate-desktop/

Oh I see, @cwendling. I'll try that as soon as I can. I haven't had much time to test it, but I think I'll be able to do it till the end of the week max. Will try to do my best. Thanks for your guys' patience 👍

caiodev:
Hello guys. Just to update you on my attempts to test the solution branch, I really could not build it because the version of MATE I'm using is not compatible with the required one and I got this error:

```
Package dependency requirement 'mate-desktop-2.0 >= 1.27.1' could not be satisfied.
Package 'mate-desktop-2.0' has version '1.26.2', required version is '>= 1.27.1'
```

Screenshot at 2025-02-03 01-53-08

I downloaded the latest version of MATE from their Github page as @cwendling  suggested but since I own 2 PCs running Ubuntu MATE 24.04.1 and I always try to test anything MATE related (New releases etc) on baremetal because I think it is a more accurate environment.

When I was about to start building it, I saw reports of some inconsistencies on the latest MATE vesion and since these PCs are shared by the whole family, I decided I couldn't go any further because of that and because I might mess with something that will take me some time to fix making at least one of those PCs unavailable for use for some time ):

Since this PR has been open for 1 month now and @raveit65 has already tested it, imho we could go with his report on it but that's you guys' call.

Finally, I'd like to thank you guys once again for your patience, for submitting/testing this fix and for all you guys do here for the community 🤝️

cwendling:
@caiodev if you're not afraid of testing binaries from the web, you're in luck today: I just built a patched Ubuntu 24.04 package with the changes form this PR applied: [marco_1.26.2-4.1_amd64.zip](https://github.com/user-attachments/files/18656627/marco_1.26.2-4.1_amd64.zip)

Issue #788, opened 2025-03-16
Design of alt-tab popup with css


luigifab:
I've just updated my CSS theme for XFCE's AltTab (`#xfwm-tabwin`), and I see that Mate's AltTab has a kind of border-radius with a white background (when background window is mate-terminal or geany). See:

![Image](https://github.com/user-attachments/assets/98a36b08-bf16-4ce7-9872-4b06f281635e)

What are the CSS nodes for it?
Is there an id or a name?

My idea is to remove the border-radius.
Thanks!

_Marco 1.26.2 + Mate 1.26.2 / Debian Testing_

lukefromdc:
I do not know myself, and this is the sort of thing it might not be possible to run GtkInspector on.
In the source code a string search for `gtk_widget_set_name` comes up empty but we do have
```

      gtk_style_context_add_class (gtk_widget_get_style_context (GTK_WIDGET (popup->window)),
                                   GTK_STYLE_CLASS_OSD);
```
adding the OSD style class to the tab popup

luigifab:
Yes, it is, this is the frame border from my theme:

```css
frame > border {
	border: 1px solid @frame_border;
	border-radius: 2.5px;
	box-shadow:
		inset -1px -1px alpha(@entry_bg, 0.27),
		inset 1px 1px alpha(@frame_border, 0.17);
}
```

PR #796, opened 2025-06-02
Add class for AltTab

luigifab:
Help #788 and Reopen #789
Tested by rebuilding MATE 1.26 Debian source package.

This allow to theme AltTab / TabWin popup, I found these CSS nodes:
```css
frame.mate-tabwin { border:1px solid orange; }
frame.mate-tabwin > border { border:1px solid yellow; }
frame.mate-tabwin box { }
frame.mate-tabwin box grid { }
frame.mate-tabwin box label { }

/* with .osd when compositor is enabled */
```

image

### DIFF
--- a/src/ui/tabpopup.c
+++ b/src/ui/tabpopup.c
@@ -403,6 +403,7 @@ meta_ui_tab_popup_new (const MetaTabEntry *entries,
   gtk_container_set_border_width (GTK_CONTAINER (grid), 16);
   gtk_container_add (GTK_CONTAINER (popup->window), frame);
   gtk_container_add (GTK_CONTAINER (frame), vbox);
+  gtk_style_context_add_class (gtk_widget_get_style_context (frame), "mate-tabwin");
 
   gtk_box_pack_start (GTK_BOX (vbox), grid, TRUE, TRUE, 0);
 


lukefromdc (commented):
Builds without issue. Was not able to test run as I am in the middle of a mate-wayland session and don't want to close all applications and terminate it right now.

This however should be guaranteed to work and is very simple. To use it themes can style the "mate-tabwin" class

src/ui/tabpopup.c
```diff
@@ -403,6 +403,7 @@ meta_ui_tab_popup_new (const MetaTabEntry *entries,
   gtk_container_set_border_width (GTK_CONTAINER (grid), 16);
   gtk_container_add (GTK_CONTAINER (popup->window), frame);
   gtk_container_add (GTK_CONTAINER (frame), vbox);
+  gtk_style_context_add_class (gtk_widget_get_style_context (frame), "mate-tabwin");
```
vkareh, line 406: nitpick: why not call the class `tabpopup` or `tab-popup` (or even `tab-popup-frame`)? It would match the rest of the code (which refers to this window as such).

vkareh: Sure, but:
```
~/dev $ git -C xfce/xfwm4 grep tabwin | wc -l
351
~/dev $ git -C mate/marco grep tabwin | wc -l
0
```
xfwm uses the `tabwin` name heavily in their window manager, most of the code in `tabwin.c`. In contrast, marco has `tabpopup.c` and uses the name `tabpopup` throughout.

vkareh (changes requested):

PR #798, opened 2025-07-17
Fix GTK deprecations

vkareh:
This fixes all remaining GTK deprecation warnings. No functionality changes.


lukefromdc (approved):
A build of this against git master worked fine and there were no issues in a test x11 session under marco. The marco-window-demo program also works fine with this applied

PR #801, opened 2025-08-07
Add dynamic workspaces support

### LINKED ISSUE #406: [feature request] dymanic workspaces
Dynamic workspaces is a really nice feature from gnome 3: I'm suggesting to implement it.

This feature does:
- create a new (last) workspace, if there is no more an empty one;
- delete any empty workspace, if there is more than one empty.

Personally, I find it more useful then keeping a lot of empty workspaces or opening the workspace switcher to add new ones time by time.

vkareh:
Adds GNOME3-style dynamic workspace management. Workspaces are created when no empty ones exist and removed when multiple empty workspaces are present, maintaining exactly one empty workspace at the end.

Fixes #406

### DIFF
--- a/src/core/prefs.c
+++ b/src/core/prefs.c
@@ -105,6 +105,7 @@ static gboolean raise_on_click = TRUE;
 static gboolean attach_modal_dialogs = FALSE;
 static char* current_theme = NULL;
 static int num_workspaces = 4;
+static gboolean dynamic_workspaces = FALSE;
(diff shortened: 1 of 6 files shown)


lukefromdc (approved):
This works quite well, and even the workspace switcher in the panel picks up changes in the number of workspaces showing. For the moment it is a non-default (as it should be being a behavior change) only exposed by gsettings but a later PR could make this togglable from the control center windows applet and/or one for mate-panel to toggle it from the switcher.

PR #804, opened 2025-08-14
theme: Extend button clickable area to edge of screen

vkareh:
Apply Fitt's law to corner window buttons by extending the clickable area all the way to the edge of the screen on maximized/tiled windows. This means that a user moving the pointer all the way to the top-right corner and clicking, will close the maximized (or right-tiled) window. The same applies for top-left on maximized and left-tiled windows.

Fixes https://github.com/mate-desktop/mate-themes/issues/315

### DIFF
--- a/src/ui/theme.c
+++ b/src/ui/theme.c
@@ -1004,7 +1004,17 @@ meta_frame_layout_calc_geometry (const MetaFrameLayout  *layout,
           rect->clickable.height = button_height;
 
           if (i == n_right - 1)
-            rect->clickable.width += layout->right_titlebar_edge + layout->right_width + layout->button_border.right;
+            {
+              /* Extend clickable area to the right edge */
+              rect->clickable.width += layout->right_titlebar_edge + layout->right_width + layout->button_border.right;
+
+              /* For maximized and right-tiled windows, extend clickable area to the top edge (Fitts's law)*/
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
This works, and it seems to work with ALL themes. In my UbuntuStudio-Legacy theme on the wayland session, the buttons work as they are placed and the top right corner (outside the button area) does not close the window. Not something that ever bothered me, but it provided a test theme where a maximized window cannot normally be closed by clicking the top right corner.

With this applied, both marco and compiz-reloaded(which is normally built to use marco for theming) would close a maximized SSD window for a top right corner click on my normal theme or on (tested) Yaru-green, which I used as a test theme since this was said to work by the theme's button placement on menta.

If anyone wants to extend this to the buttons on the left case (including menu button on the left in the most common layout) I have no problem with that, with leaving that for another PR, or not doing so. I don't know Fitt's law in general and thus cannot judge what to do with the left corner

davidhedlund:
> If anyone wants to extend this to the buttons on the left case (including menu button on the left in the most common layout) I have no problem with that, with leaving that for another PR, or not doing so. I don't know Fitt's law in general and thus cannot judge what to do with the left corner

@vkareh Your fix will help so many users, thank you so much! I'm curious—can this same approach be adapted to tackle the related top-left fitts' law issue as well?:

* https://github.com/mate-desktop/mate-themes/issues/318

PR #806, opened 2025-08-29
Add area screenshot support

vkareh:
This commit fixes a few issues in marco that prevent it from having a keybinding for area screenshot (e.g. `mate-screenshot --area`). With those fixed, I added the actual keybinding for it, defaulting to `<Shift>Print`.


lukefromdc (approved):
A rebase of this against newly updated master builds clean (no new warnings) and works as intended. Note that in testing/using, the initial reaction to the shift-print key combo is the cursor changes to the cross used to select the area. You use this to select the area to copy, and mate-screenshot comes up with that area selected when you release the mouse.

I'll have to see if grim has implemented anything like this for use in the wayfire session, this can be quite handy. Will merge.

PR #809, opened 2025-09-03
theme: Fix left-side Fitt's Law

vkareh:
When calculating the clickable area on the left-most button of a maximized window, we need to take into account the layout components, not just the button size.

Fixes mate-desktop/mate-themes#318


vkareh:
@davidhedlund - I've fixed the leftmost button clickable area now (and cleaned up the code for the rightmost one a bit too).

PR #813, opened 2025-09-22
ci: Trigger full release on new tag

vkareh:
Whenever there's a new v* tag, GitHub will use distcheck to generate tarballs, create a proper release, and trigger the mate-desktop.org webhook.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -98,3 +98,15 @@ jobs:
 
       - name: Build the source code
         run: .github/workflows/builds.sh ${{ matrix.build-system }}
+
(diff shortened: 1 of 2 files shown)


vkareh:
@lukefromdc I don't know if this will work, but I pulled this stuff from when we used Travis, specifically commit https://github.com/mate-desktop/marco/commit/b3f039b66b058886fadc1c7b124a900625889924 where the release job notifies the release.mate-desktop.org API.

I suspect that this is the missing piece to get tarballs into pub.mate-desktop.org, but I don't know for sure (I guess unless we merge this and tag a new version).

If you have other ideas, I'm open...

PR #814, opened 2025-10-02
screen: Add dynamic HiDPI scale change support

vkareh:
This allows marco to respond to HiDPI scale changes without requiring a restart. When the scale changes, marco now automatically updates all icons (windows, xprops, alt+tab, etc.) and mouse cursor to be at the correct scale factor.

Window decorations are handled through an internal GDK signal triggered from XSettings changes, and so already work out of the box.

### DIFF
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -65,6 +65,9 @@ static void set_workspace_names    (MetaScreen *screen);
 static void prefs_changed_callback (MetaPreference pref,
                                     gpointer       data);
 
+static void on_monitors_changed    (GdkScreen      *gdk_screen,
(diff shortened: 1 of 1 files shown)


vkareh:
> Question: does the `g_slist_free (windows);` invocation also free the` tmp` GSList which is set equal to it? I don't know enough about GSList to answer that myself

No, freeing `windows` doesn't free `tmp`. However, `tmp` doesn't itself hold a different list - it's a pointer to the same list in memory. Freeing `windows` is sufficient. At that point, `tmp` is just a local dangling pointer (it goes away after the function context is gone).

PR #819, opened 2025-12-21
alt+tab popup: add preference to mix minimized and unminimized windows

ericek111:
Equivalent of `org.cinnamon.alttab-minimized-aware` in Cinnamon desktop.
If true (default), minimized windows will be put last in the alt+tab list, otherwise they'll be mixed, allowing the user to minimize a window and recall it quickly with Alt+Tab.

In the past few weeks, this has been an ever-increasing pain in the butt for me. This doesn't change the current behavior. It merely adds an option for the user to choose their preference. The icons of minimized windows will continue to be rendered slightly transparent.


ericek111:
I'm testing out KDE Plasma on my tablet and this behavior is the default. Please, consider merging this, I'm going crazy on my other systems.

vkareh:
@ericek111 A few months ago I tried addressing this in a more generic way, rather than have a single gsettings that does one thing. Instead I added gsettings that change the alt+tab position for minimized and urgent windows. Can you try that and see if it solves it for you? Here it is: https://github.com/mate-desktop/marco/pull/833

PR #826, opened 2026-02-09
display: Use global MRU list for switch-windows-all

### LINKED ISSUE #286: switch-windows-all duplicates omnipresent windows in window switcher dialog
Running Mate 1.16.0 on ArchLinux
Running Marco 1.16.0 GTK3

When using the keybind for `switch-windows` Marco performs as expected when using omnipresent windows. The window shows up on each desktop and is only present in the Window Switcher once.

switcher_normal

Note that in this screenshot, the firefox window is set to appear on all workspaces. It works as expected and only shows up in the switcher once.

However, when `switch-windows-all`  used, the following bug happens.

switcher_bug

In this case the window I have used to showcase the bug is Caja, but this is reliably reproducable with (I believe) any application. Notice how even though my panel shows only one open instance of Caja, the switcher dialog now shows four entries for Caja (1 for each virtual workspace). Selecting any of these entries brings me to my single open instance of Caja.

To reproduce:
1. Bind `switch-windows-all` (or bind to seperate keybinds)
2. Create a new window of any application.
3. Mark the window as omnipresent (either by keybind or via `Always on Visible Workspace` menu itme)
4. Activate the window-switcher.

You should see duplicate entries for your omnipresent window equal to the number of virtual desktops. Multiple windows can be marked as omnipresent, and all will show up duplicated in the window switcher. Ex: Mark Caja and Firefox to `Always on Visible Workspace` and Alt-Tab, you will see four Caja and four Firefox entries.

I noticed this issue as well in Marco 1.14.0 but did not get around to filing until now. I did not use MATE before 1.14 and cannot comment on earlier versions of Marco.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38041243-switch-windows-all-duplicates-omnipresent-windows-in-window-switcher-dialog?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
When using switch-windows-all to cycle through windows across all workspaces, marco was grouping windows by workspace rather than using a global MRU (most-recently-used) list. This caused several issues:

- Sticky windows appeared multiple times, once per workspace
- Window order was confusing (since they were grouped by workspace number)
- Behavior was different from other window managers

Fixed by collecting all windows, sorting them by recency, then building the tab list from this global list.

Fixes #286

### DIFF
--- a/src/core/display.c
+++ b/src/core/display.c
@@ -4705,72 +4705,80 @@ find_tab_backward (MetaDisplay   *display,
   return NULL;
 }
 
+static int
+mru_cmp (gconstpointer a,
+         gconstpointer b)
+{
+  guint32 time_a, time_b;
+
+  time_a = meta_window_get_user_time ((MetaWindow *)a);
+  time_b = meta_window_get_user_time ((MetaWindow *)b);
+
+  if (time_a > time_b)
+    return -1;
+  else if (time_a < time_b)
+    return 1;
+  else
+    return 0;
+}
+
 GList*
 meta_display_get_tab_list (MetaDisplay   *display,
                            MetaTabList    type,
                            MetaScreen    *screen,
                            MetaWorkspace *active_workspace)
 {
-  GList *tab_list, *workspace_list, *l, link;
-  MetaWorkspace *workspace;
+  GList *tab_list, *mru_list, *l;
 
   g_return_val_if_fail (active_workspace != NULL, NULL);
 
+  /* Build MRU list based on whether we use global or per-workspace */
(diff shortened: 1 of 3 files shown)


mbkma:
ah okay just found out, you can put for example `<Control>a` in there

mbkma (approved):
It builds and runs as intended, code looks good

PR #829, opened 2026-03-20
window: Exclude invisible borders from wireframe rectangle

### LINKED ISSUE #532: Empty space around windows
Take a window with the mouse (either by titlebar or with Alt pressed) and try to move it. A frame around the window, slightly bigger than the window itself, appears to allow you position the window. Move the frame to match a screen corner.

#### Expected behaviour
After the movement finished, the window should be placed exactly in the corner.

#### Actual behaviour
There is a wide band between screen edges and the window.

Listed above.

#### MATE general version
1.22

#### Package version
marco-1.22.2-1.fc30.x86_64

#### Linux Distribution
Fedora 30

vkareh:
The wireframe drawn in reduced-resources mode included invisible resize borders, making it larger than the visible window. This removes the invisible borders from the xor rect calculation.

Fixes #532

### DIFF
--- a/src/core/screen.c
+++ b/src/core/screen.c
@@ -529,7 +529,7 @@ meta_screen_new (MetaDisplay *display,
     gc_values.line_width = META_WIREFRAME_XOR_LINE_WIDTH;
     value_mask |= GCLineWidth;
 
-    font_info = XLoadQueryFont (screen->display->xdisplay, "fixed");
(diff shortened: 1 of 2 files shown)


mbkma (approved):
Looks good to me. Excluding the invisible borders from the wireframe rectangle is the right way to make the resize outline match the visible window bounds again, and the change stays nicely contained to the geometry calculation.

PR #830, opened 2026-03-20
frame: Preserve client position when removing decorations

### LINKED ISSUE #534: After resuming window from tray, window position shift for qt application
#### Expected behaviour
After resuming window from tray, window size and position has to be the same as before closing to system tray for all applications.

#### Actual behaviour
After resuming window from tray, window position shifted to up and left for most modern QT applications. This behavior I can reproduced for another DE (tested for XFCE and LXQT) after replacement window manager to **marco** or **mutter** (WM for gmone 3)

Install vlc or qps or telegram. Adjust them to completely close (or hide) to system tray (I mean, after closing app, it's window has to completely disappear from desktop and panel too). Close (or hide) application to tray and resume it: new window position has shifted to up and a bit left (attached screenshots demonstrated it).

#### MATE general version
1.20.0+5 (debian stable) or 1.22.0+1 (debian sid)

#### Package version
1.20.3-1 (debian stable) or 1.22.2-1 (debian sid)

#### Linux Distribution
Debian stable (Buster) or debian sid

1 Before cls to tray
2 Closed to tray
3 Resumed from tray

vkareh:
Windows using StaticGravity (like Qt apps) would shift up and left by the visible frame border size each time they were hidden to system tray and restored. The position drift compounded with each cycle. This changes those windows to use the client's actual root position.

Fixes #534

### DIFF
--- a/src/core/frame.c
+++ b/src/core/frame.c
@@ -170,6 +170,7 @@ meta_window_destroy_frame (MetaWindow *window)
 {
   MetaFrame *frame;
   MetaFrameBorders borders;
+  int offset_x, offset_y;
 
   if (window->frame == NULL)
     return;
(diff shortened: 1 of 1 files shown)


mbkma (approved):
Good fix. Preserving the client position for StaticGravity windows when decorations are removed matches the way those windows expect their coordinates to behave, and keeping the adjustment limited to that case avoids changing the existing behavior for the other gravity modes.

PR #833, opened 2026-05-05
display: Add settings for Alt+Tab window placement

### LINKED ISSUE #824: Alt+Tab changes workspace somewhat randomly
This is a long standing and annoying bug (to me at least. I hope it's not intentional, anyway!)

Sometimes, alt+tab does not cycle through the windows of a given workspace.  It switches to another workspace

It happens Esp. 
- when there are multiple windows of the same program (vscode, firefox), 
- but also sometimes when i open a program, place it somewhere with Ctrl+Alt+left/right then go back to initial desktop, then alt/tab -> goes to the program's workspace
- and also sometimes more randomly (it seems, and maybe when there is a notification)

This bug (or I hope it's one) has existed for many years and is really annoying, as it breaks concentration (when switching windows rapidly and landing on a random workspace, thinking “oh no, where am I now?”, and losing focus), and I couldn’t find it in the MATE bug tracker.

My marco version is 1.26.2 (the one in debian 13).

Thanks!

vkareh:
Add user settings to control how minimized and urgent windows are ordered during Alt+Tab.

1. alt-tab-minimized-placement:
  - 'mru': Mix minimized windows in MRU order
  - 'end': Show minimized windows last (current behavior, default)
  - 'hidden': Exclude minimized windows from Alt+Tab

2. alt-tab-urgent-placement:
  - 'start': Show all urgent windows first (current behavior, default)
  - 'mru': Mix urgent windows in MRU order

Fixes #824
Closes #819


mbkma (commented):
The refactor looks clean overall, but I think this changes one piece of existing behavior along the way: urgent windows now seem to get pulled to the front even when they are on the current workspace, whereas the old code only gave that treatment to urgent windows from other workspaces. If that change was intentional then the new structure is fine, but if not it may be worth restoring the workspace check in the urgent-window path.

vkareh:
I thought the existing code treated urgent windows the same regardless of the workspace. But if that's not the case, then this change will fix it.

mbkma:
I think in general it is better to wait for an approval before merging.

vkareh:
> I think in general it is better to wait for an approval before merging.

Oh shoot, you're right about that, sorry.

PR #834, opened 2026-05-15
window: focus modal dialogs whose ancestor is focused

### LINKED ISSUE #784: Modal windows don't get focus when displayes
#### Expected behaviour

Modal windows should get the focus, when displayed.

#### Actual behaviour

The windows are displayed, but they don't get the focus, so one needs to manually click on them.

I'm not 100% sure that this triggers only on modal windows, but I think that's the case.

There are several programs triggering this bug. One of them is Visual Studio Code. Open a text file, then invoke "Save as": the modal window is displayed, but not in focus.

#### MATE general version

mate-desktop 1.26.2-1.1build3.

#### Package version

1.26.2-4build4

#### Linux Distribution

Ubuntu MATE 24.04.1

#### Link to bugreport of your Distribution (requirement)

https://bugs.launchpad.net/ubuntu/+source/marco/+bug/2080515

vkareh:
Fixes #784.

Commit 6ea23df ("window: do not unfocus on new window") imported the Metacity bf17c79 / Mutter 998d921d behavior for GNOME Bugzilla #773210: when a new transient maps without taking focus, keep focus on the parent so GTK popup and completion windows and VLC fullscreen controls do not steal keyboard focus.

That blanket change also affected modal dialogs whose parent already has focus. Electron and GTK file chooser dialogs can reach meta_window_show() with takes_focus_on_map false because their _NET_WM_USER_TIME is missing or stale. Since 6ea23df, such dialogs are mapped above the parent but keyboard focus stays on the parent, so typing goes to the wrong window.

Keep the bf17c79-equivalent behavior for non-modal transients, but when the new window is a META_WINDOW_MODAL_DIALOG whose ancestor is the current focus window, force takes_focus_on_map so the blocking dialog receives keyboard focus immediately.

### DIFF
--- a/src/core/window.c
+++ b/src/core/window.c
@@ -2158,6 +2158,14 @@ meta_window_show (MetaWindow *window)
           if (!window->placed)
             window->denied_focus_and_not_transient = TRUE;
         }
+      else if (window->type == META_WINDOW_MODAL_DIALOG)
+        {
+          /* Keep the non-modal transient behavior from 6ea23df for GTK
+           * popups and VLC controls, but focus modal dialogs that block
+           * the focused ancestor.
+           */
+          takes_focus_on_map = TRUE;
+        }
     }
 
   if (!window->placed)


mbkma (approved):
Nice targeted fix. Limiting the special handling to modal dialogs whose ancestor is already focused preserves the earlier transient-focus behavior while still fixing the cases where the modal should clearly take focus immediately.

PR #839, opened 2026-08-11
Always include config.h before util.h

vkareh:
The util.h file guards verbose output through `WITH_VERBOSE_MODE`, which is defined in config.h. This means that verbose mode would have never worked on some of these files unless config.h was included first.

All other files with verbose output calls already include config.h.


lukefromdc:
How do I test this?

vkareh:
There's nothing to test really, this just prevents `meta_verbose` calls from being silently dropped by the compiler when someone (usually me :sweat_smile:) is debugging marco.

lukefromdc (approved):
I just got clean builds with this both in meson and in autotools, no problems detected that I could discern

